PR #4, opened 2019-01-07
import gcp auth plugin in kubeutils

ilackarms:
so you don't have to keep importing it all over the darn place :slightly_smiling_face:


ilackarms:
closing this in lieu of a discussion on the right place to put this import

PR #6, opened 2019-01-18
remove solo-kit related utils

yuval-k:


yuval-k:
cloud build passed, but i removed its permissions mid build so it doesnt show.

PR #8, opened 2019-01-31
test utils

yuval-k:


testutils/cli/testutil.go
```diff
@@ -0,0 +1,98 @@
...
+
+	donec := make(chan struct{})
+	go func() {
+		defer GinkgoRecover()
+		defer close(donec)
+
+		userinput(&Console{Console: c})
+	}()
+
+	//	time.Sleep(time.Hour)
```
rickducott: This should probably be removed

PR #11, opened 2019-02-21
copied docker code

EItanya:


docker/docker_test.go
```diff
@@ -0,0 +1,62 @@
...
+	"os/exec"
+)
+
+const (
+	validImage   = "soloio/gloo:0.7.0"
+	invalidImage = "soloio/gloo:0.1.0"
+)
+
+var _ = BeforeSuite(func() {
+	exec.Command("docker", "image", "prune", "-af").Run()
```
yuval-k: can you only prune images you pulled?
i have some cached images that I want to stay cached.

docker/exec.go
```diff
@@ -0,0 +1,67 @@
...
+)
+
+
+var (
+	logger *zap.SugaredLogger
+)
+
+func init() {
+	var err error
+	devLogger, err := zap.NewDevelopment()
```
yuval-k: i dont think go-utils should initialize logging. especially not a development logger.
the main problem with dev log, is that DPanic will panic if we use it in prod.
but in generated i think we should avoid non-main packages initializing logging

Ideally the package should receive the logger from the context, or as a parameter.

PR #12, opened 2019-02-25
Changelog

EItanya:


changelogutils/github.go
```diff
@@ -0,0 +1,68 @@
...
+			Bytes: byt,
+			Filename: file.GetName(),
+		})
+	}
+
+	changelog, err := BytesToChangelog(rawChangelogFiles)
+	if err != nil {
+		return false, err
+	}
+	fmt.Println(changelog)
```
rickducott: Remove println

githubutils/githubutils_test.go
```diff
@@ -0,0 +1,48 @@
...
+)
+
+var _ = Describe("github utils", func() {
+	var (
+		client   *github.Client
+		ctx      = context.Background()
+		owner    = "solo-io"
+		reponame = "testrepo"
+		version = "v0.0.16"
+		ref = "master"
```
rickducott: Might make sense to use ref = "v0.0.16" too or some commit hash, don't want the test to break because master changes in the future

changelogutils/changelog.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+type RawChangelogFile struct {
+	Filename string
+	Bytes []byte
+}
+
+const (
+	ChangelogDirectory = "changelog"
+	Master = "master"
```
rickducott: how is this used?

changelogutils/changelog.go
```diff
@@ -0,0 +1,68 @@
+package changelogutils
+
+import (
+	"context"
+	"github.com/ghodss/yaml"
+	"github.com/solo-io/go-utils/githubutils"
+)
+
+type ChangelogEntryType int
```
rickducott: Should probably be string constant but we can figure it out in a future PR

changelogutils/github.go
```diff
@@ -0,0 +1,68 @@
...
+		}
+	}
+	if proposedVersion == "" {
+		return "", errors.Errorf("No version greater than %s found", latestTag)
+	}
+	return proposedVersion, err
+}
+
+func ValidateProposedChangelogTag(ctx context.Context, client *github.Client, owner, repo, ref string) (bool, error) {
+	proposedTag, err := GetProposedTagFromGit(ctx, client, owner, repo, ref)
```
rickducott: The flow here is a bit unexpected, the test for this file first calls getProposed, then calls validateProposed which internally calls getProposed again. I'd suggest we slightly change it:
* GetProposedTag -> takes a ref, figures out which, if any, tag / directory is the proposed release tag (proposed = higher semver version than current). returns error if there are 0 or multiple proposed
* Validate -> takes a tag and the current ref, checks the directory contents: a) makes sure they are valid and parseable, b) makes sure number has been properly incremented based on whether there is a breaking change

changelogutils/changelog.go
```diff
@@ -0,0 +1,95 @@
...
+	Bytes []byte
+}
+
+const (
+	ChangelogDirectory = "changelog"
+	Master = "master"
+)
+
+// Should return the last released version
+// Executes git commands, so this won't work if running from an unzipped archive of the code.
```
rickducott: This comment is out of date

githubutils/githubutils_test.go
rickducott: Also gofmt

PR #15, opened 2019-02-26
Changelog validation

rickducott:


changelogutils/changelog.go
```diff
@@ -74,16 +76,74 @@ func GetProposedTag(fs afero.Fs, latestTag, changelogParentPath string) (string,
...
-	if err := yaml.Unmarshal(bytes, changelog); err != nil {
-		return nil, errors.Wrap(err, "failed parsing changelog file")
+	if err := yaml.Unmarshal(bytes, &changelog); err != nil {
+		return nil, errors.Errorf("File %s is not a valid changelog file", path)
 	}
 
 	return &changelog, nil
 }
+
+func VersionToString(v *versionutils.Version) string {
```
EItanya: nit: make this a property of Version instead so it also the fmt stringer interface
func (v *Version) String() string

changelogutils/changelog.go
```diff
@@ -74,16 +76,74 @@ func GetProposedTag(fs afero.Fs, latestTag, changelogParentPath string) (string,
...
+	return fmt.Sprintf("v%d.%d.%d", v.Major, v.Minor, v.Patch)
+}
+
+func hasBreakingChange(file *ChangelogFile) bool {
```
EItanya: same as above

versionutils/version.go
```diff
@@ -93,3 +93,33 @@ func MatchesRegex(tag string) bool {
 	regex := regexp.MustCompile("(v[0-9]+[.][0-9]+[.][0-9]+$)")
 	return regex.MatchString(tag)
 }
+
+func IncrementVersion(v *Version, breakingChange bool) *Version {
```
EItanya: Could make patch, minor, and major functions so this package could be used elsewhere, not just by changelog

changelogutils/changelog.go
```diff
@@ -15,20 +14,32 @@ import (
 type ChangelogEntry struct {
 	Type        ChangelogEntryType
 	Description string
+	IssueLink   string
```
EItanya: I like this a lot

PR #22, opened 2019-02-27
ref has changelog test

EItanya:


changelogutils/changelog.go
```diff
@@ -175,3 +176,20 @@ func ComputeChangelog(fs afero.Fs, latestTag, proposedTag, changelogParentPath s
...
+func RefHasChangelog(ctx context.Context, client *github.Client, owner, repo, sha string) (bool, error) {
+	opts := &github.RepositoryContentGetOptions{
+		Ref: sha,
+	}
+
+	_, branchRepoChangelog, branchResponse, err := client.Repositories.GetContents(ctx, owner, repo, ChangelogDirectory, opts)
+	if err == nil && len(branchRepoChangelog) > 0 {
+		return true, nil
+	} else {
+		if branchResponse.StatusCode != 404 {
```
rickducott: What status code are we expecting here? I thought 404 is right

rickducott: Nvm, I understand

EItanya: basically 404 isn't actually an error. So i'm not surfacing an error
404 means the changelog directory doesn't exist

PR #27, opened 2019-02-28
Fix changelog pr

rickducott:


rickducott:
Filed #28 to track the real fix

PR #37, opened 2019-03-03
Upload release asset cli

rickducott:
Migrating repositories to `UploadReleaseAssetCli`, distributed in `go-utils/githubutils/cli.go`. 

Convention changes:
* With the `SkipAlreadyExists: true` flag, this script will not error if some / all of the artifacts were already uploaded (does simplistic checking on name only). This enables re-running release builds without needing to clear state, when a release build fails. 
* Use this CLI instead of shell script. Script will be removed from repos. 
* Move away from `release` target, create `upload-github-release-assets` make target instead. 
* Target is always run, logic to determine if this is a release is in the CLI. Make target should be: 
```.PHONY: upload-github-release-assets
upload-github-release-assets:
	go run upload_github_release_assets.go
```

This assumes the release artifacts have already been built. To enforce this, you could add the related make targets as a dependency here, though then we're duplicating the dependencies of targets.

### DIFF
--- a/githubutils/README.md
+++ b/githubutils/README.md
@@ -1,0 +1,65 @@
+## Uploading Release Assets to Github
+
+To upload release assets to Github, follow these steps (requires go-utils 0.2.10+). 
+
+### Create a Go script
+
+Create a script `upload_github_release_asset.go`, like this: 
+
+```go
+package main
+
+import (
+	"github.com/solo-io/go-utils/githubutils"
+)
+
+func main() {
+	assets := make([]githubutils.ReleaseAssetSpec, 2)
+	assets[0] = githubutils.ReleaseAssetSpec{
+		Name: "hello",
+		ParentPath: "_output",
+		UploadSHA: true,
+	}
+	assets[1] = githubutils.ReleaseAssetSpec{
+		Name: "my-resource.yaml",
+		ParentPath: "namespace",
+	}
+	spec := githubutils.UploadReleaseAssetSpec{
+		Owner: "solo-io",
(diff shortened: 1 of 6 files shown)


rickducott:
Tested supergloo for fix to #40, confirmed with this PR: https://github.com/solo-io/solo-docs/pull/85/files

Ran in bash: 
```
export GITHUB_TOKEN=...
export TAGGED_VERSION=v0.2.0
make push-docs
```

githubutils/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+```go
+package main
+
+import (
+	"github.com/solo-io/go-utils/githubutils"
+)
+
+func main() {
+	assets := make([]githubutils.ReleaseAssetSpec, 2)
```
EItanya: Could this possibly be done with a variadic function, or something like that?

githubutils/upload_release_asset.go
```diff
@@ -0,0 +1,116 @@
...
+	defer file.Close()
+	if err != nil {
+		log.Fatalf("Error reading file %s: %s", path, err.Error())
+	}
+	opts := &github.UploadOptions{
+		Name: name,
+	}
+	_, _, err = client.Repositories.UploadReleaseAsset(ctx, spec.Owner, spec.Repo, release.GetID(), opts, file)
+	if err != nil {
+		log.Fatalf("Error uploading assets. Error was: %s", err.Error())
```
EItanya: I think we should use context logging with contextutils wherever possible

PR #52, opened 2019-03-06
Dependency changelog

rickducott:


rickducott:
See https://github.com/solo-io/testrepo/pull/28 for some ideas on how this might work.

PR #58, opened 2019-03-07
util function to wait for a crd to be registered

ilackarms:
extracted from supergloo

i plan to use this in supergloo as well as solo-kit


kubeutils/wait_crd_test.go
```diff
@@ -0,0 +1,44 @@
...
+	apiexts "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
+	"k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var _ = Describe("WaitCrd", func() {
+	var (
+		api     apiexts.Interface
+		crdName = "testing"
+	)
+	BeforeEach(func() {
```
marcogschmidt: Nit: since it is a simgle test you could put everything in the `It` closure.

PR #69, opened 2019-03-12
Cleanup3

rickducott:


rickducott:
Can work on this refactor on the changelog-refactor branch, but removing from master so it won't be used

PR #73, opened 2019-03-14
cluster lock timeout

EItanya:


testutils/clusterlock/clusterlock_test.go
```diff
@@ -71,6 +71,21 @@ var _ = Describe("cluster lock test", func() {
 		Expect(err).NotTo(HaveOccurred())
 		Expect(lock2.AcquireLock(retry.Delay(time.Millisecond))).To(HaveOccurred())
 	})
+
+	FIt("Take back timed out lock", func() {
```
ilackarms: It

testutils/clusterlock/main.go
```diff
@@ -78,10 +84,27 @@ func (t *TestClusterLocker) AcquireLock(opts ...retry.Option) error {
 			}
 
 			if cfgMap.Annotations == nil || len(cfgMap.Annotations) == 0 {
```
ilackarms: this can be simplified to just `len(cfgMapp.Annotations) == 0`

PR #74, opened 2019-03-14
Not lock owner

EItanya:


testutils/clusterlock/clusterlock_test.go
```diff
@@ -87,6 +87,15 @@ var _ = Describe("cluster lock test", func() {
 		Expect(lock2.AcquireLock()).NotTo(HaveOccurred())
 		Expect(lock2.ReleaseLock()).NotTo(HaveOccurred())
 	})
+
+	It("only the user holding the lock can release it", func() {
```
rickducott: What about a test that a different user trying to release does throw an error?

EItanya: That's what this test does

PR #78, opened 2019-03-21
clusterlock-heartbeat

EItanya:


testutils/clusterlock/main.go
```diff
@@ -60,96 +64,153 @@ type TestClusterLocker struct {
...
+
+
+	if options.Namespace == "" {
+		options.Namespace = LockDefaultNamespace
+	}
+	if options.Context == nil {
+		options.Context = context.Background()
 	}
-	_, err := clientset.CoreV1().ConfigMaps(namespace).Create(defaultConfigMap)
+	buildId := options.IdPrefix + uuid.New().String()
```
ilackarms: can we correlate build id with something from the actual build?

EItanya: It needs to be a uuid, but an option for a prefix can be passed in which can be used to ID

PR #89, opened 2019-03-28
Feature/resolves issue

rickducott:


mitchdraft:
@rickducott Looks good, but I wonder, since this is defaulting to true, and a user will need to explicitly set it to false in order to prevent the issue closing, would it be any better to let the user leave a message saying why this does not close the issue?
The field could become something like `keepIssueOpen: "Still need to add documentation"`

PR #96, opened 2019-04-01
Testutils revamp

EItanya:
BOT NOTES: 
resolves https://github.com/solo-io/go-utils/issues/95


surveyutils/input.go
```diff
@@ -0,0 +1,221 @@
...
+	return GetBoolInputDefault(msg, value, false)
+}
+
+func GetBoolInputDefault(msg string, value *bool, defaultValue bool) error {
+	var strValue string
+	defaultValueStr := "N"
+	if defaultValue {
+		defaultValueStr = "y"
+	}
+	prompt := &survey.Input{Message: msg + " [y/N]: ", Default: defaultValueStr}
```
yuval-k: upper case N implies that it is the default. make it lower case?
or make the y upper case?

Issue #97, opened 2019-04-05
include trimmed stack trace in testutils fail handler


mitchdraft:
Since Ginkgo `Expect` failures only reference the line of the `It` block, include a stack trace so we can easily see the exact `Expect` that failed

Also trim the known stack "overhead" such as `vendor` and `test` files

This was very helpful in `glooctl` tests:
https://github.com/solo-io/gloo/pull/651/commits/824051afe05c28ae831691807de4f64bc3a0459b#diff-55e3c20e27d71c2e33a48064a1e04822R33

mitchdraft:
actually ginkgo provides this out of the box, sometimes you just need to tell it how far you are offset so it can get the right part of the stack trace
New purpose of this issue: revert the changes in 
solo-io/gloo@824051a#diff-55e3c20e27d71c2e33a48064a1e04822R33
and correct the offsets

PR #98, opened 2019-04-05
include stack trace in ginkgo fail handler

mitchdraft:
BOT NOTES: 
resolves https://github.com/solo-io/go-utils/issues/97


mitchdraft:
whoops, don't need this if you correctly set the offset in ExpectWithOffset

Issue #121, opened 2019-04-14
Add support for rendering chart/manifests from github folder


rickducott:
For instance github.com/helm/charts/stable/envoy

PR #123, opened 2019-04-14
Helm from GitHub

rickducott:
BOT NOTES: 
resolves https://github.com/solo-io/go-utils/issues/121
resolves https://github.com/solo-io/go-utils/issues/122

### DIFF
--- a/fsutils/afero.go
+++ b/fsutils/afero.go
@@ -7,7 +7,6 @@ import (
 	"github.com/spf13/afero"
 )
 
-// These two utils copied from solobot
 func SetupTemporaryFiles(fs afero.Fs) (file afero.File, dir string, err error) {
 	tmpf, err := afero.TempFile(fs, "", "tar-file-")
 	if err != nil {
--- a/fsutils/afero.go
+++ b/fsutils/afero.go
@@ -1,0 +1,43 @@
+package fsutils
+
+import (
+	"fmt"
+	"path/filepath"
+
+	"github.com/spf13/afero"
+)
+
+// These two utils copied from solobot
+func SetupTemporaryFiles(fs afero.Fs) (file afero.File, dir string, err error) {
+	tmpf, err := afero.TempFile(fs, "", "tar-file-")
+	if err != nil {
+		return nil, "", err
+	}
+
+	tmpd, err := afero.TempDir(fs, "", "tar-dir-")
+	if err != nil {
+		return nil, "", err
+	}
+	return tmpf, tmpd, err
+}
+
+func GetRepoFolder(fs afero.Fs, tmpd string) (string, error) {
+	files, err := afero.ReadDir(fs, tmpd)
+	if err != nil {
+		return "", err
+	}
+	if len(files) != 1 {
+		return "", fmt.Errorf("expected only one folder from archive tar, found (%d)", len(files))
+	}
+
+	var repoDirName string
+	for _, file := range files {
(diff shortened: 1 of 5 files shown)


fsutils/afero.go
```diff
@@ -0,0 +1,42 @@
+package fsutils
+
+import (
+	"fmt"
+	"path/filepath"
+
+	"github.com/spf13/afero"
+)
+
+func SetupTemporaryFiles(fs afero.Fs) (file afero.File, dir string, err error) {
```
mitchdraft: I think a utils lib like this is useful, but I feel this function is too specific to get any reuse outside of github archive downloads. The resources it creates `tar-dir-` and `tar-file-` are pretty specific.
How about moving this to the helm chart package where it's used or renaming it something like `SetupTemporaryFilesForGitArchive`

fsutils/afero.go
```diff
@@ -0,0 +1,42 @@
...
+	return tmpf, tmpd, err
+}
+
+func GetRepoFolder(fs afero.Fs, tmpd string) (string, error) {
+	files, err := afero.ReadDir(fs, tmpd)
+	if err != nil {
+		return "", err
+	}
+	if len(files) != 1 {
+		return "", fmt.Errorf("expected only one folder from archive tar, found (%d)", len(files))
```
mitchdraft: similar comment to above, this util has some strong assumptions, consider privatizing it or renaming

installutils/helmchart/manifests_test.go
```diff
@@ -91,4 +91,30 @@ metadata:
 
 		})
 	})
+
+	Context("load from github", func() {
+		It("works for a single chart", func() {
+			ref := GithubChartRef{
+				Ref:            "c00d740c83ec5b5439bf2f58fee110a546af7177",
```
mitchdraft: good idea locking to a ref, was going to suggest you make some dummy helm resources in this repo's test resources directory. Might still be worth doing so you can test various helm ignore rules.
Though it produces a chicken-and-egg problem, can't test against it till it merges

fsutils/afero.go
```diff
@@ -0,0 +1,42 @@
+package fsutils
```
mitchdraft: consider calling this `vfsutils` as it seems to be oriented towards virtual filesystems

fsutils/afero.go
rickducott: So I actually copied them over from solobot and didn’t want to modify or make private here so I could delete them there. Don’t mind leaving them in helm / install utils, but in the solobot case it’s for changelog verification, nothing to do with helm.

installutils/helmchart/manifests_test.go
rickducott: FWIW the helm ignore rules are designed to be used during packaging, not rendering. So I'm not sure there's a ton of value in adding testing around that. That said, I did take a look at a few of the charts and rendered manifests just to make sure things checked out.

fsutils/afero.go
rickducott: On second thought, you're right -- added a better utility and moved these to private helpers. Should be pretty easy to make use of the new function in solobot.

fsutils/afero.go
rickducott: fixed

rickducott:
Thanks @mitchdraft, done iterating -- could you take another look?

PR #126, opened 2019-04-16
Added utilities to update homebrew and gofish formulas

scranton:
Created main utility function `pkgmgmtutils.UpdateFormulas` that can be used during releases to update all three flavors of package manager style formulas for all of our CTL tooling: homebrew-core, homebrew-tap, and gofish. This utility will update the formulas and create a Pull Request against the source repo even if it's remote, e.g., github.com/homebrew/homebrew-core. Should be used in a similar form as https://github.com/solo-io/gloo/blob/a07620941be10029f13d6c00ddfa52bae141304a/ci/upload_github_release_assets.go#L50
BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/520


pkgmgmtutils/update_formulas.go
```diff
@@ -0,0 +1,386 @@
...
+		return err
+	}
+
+	byt := []byte(c)
+
+	//
+	// Update formula with new version information
+	//
+
+	byt, err = updateFormula(byt, version, versionSha, shas, fOpt)
```
EItanya: nit: `updateFormula([]byte(c)...`

pkgmgmtutils/utils.go
```diff
@@ -0,0 +1,150 @@
...
+	"bytes"
+	"errors"
+	"io/ioutil"
+	"path/filepath"
+	"regexp"
+	"strings"
+)
+
+var (
+	ErrNoSha256sFound = errors.New("did not find all needed sha256 files")
```
EItanya: Typical pattern in go as I've seen it would be:
```
errNoSha256sFound = errors.New("did not find all needed sha256 files")
var IsNoSha256sFoundErr(err error) bool {
    return err == errNoSha256sFound
}
```
Totally optional though

pkgmgmtutils/update_formula_test.go
```diff
@@ -0,0 +1,461 @@
+package pkgmgmtutils
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("package management utils", func() {
+	/*	It("can get latest release version", func() {
```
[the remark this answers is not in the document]
scranton: Keeping that test code until we integrate with other projects as that unit test has examples of all of the regex for each flavor of package manager formula.

pkgmgmtutils/utils.go
scranton: Tweaked Error code to match these conventions https://github.com/golang/go/wiki/Errors. Happy to update to better match Solo-io standards; just need to know what those are. :-)

pkgmgmtutils/update_formulas.go
```diff
@@ -0,0 +1,384 @@
...
+}
+
+type FormulaStatus struct {
+	Name    string
+	Updated bool
+	Err     error
+}
+
+func UpdateFormulas(projectRepoOwner string, projectRepoName string, parentPathSha256 string, fOpts []FormulaOptions) ([]FormulaStatus, error) {
+	//
```
rickducott: nit: Generally we can be more selective and concise with comments. For blocks like this, get rid of the leading and trailing `//` lines. And consider dropping these kinds of comment blocks altogether. I find it just as hard to read this code b/c the functions are so verbose due to the excessive spacing + comments. In the cases where the comments do add something, we could typically achieve the same readability by extracting logic to a helper function.

pkgmgmtutils/update_formulas.go
```diff
@@ -0,0 +1,384 @@
...
+				Head:                github.String(prHead),
+				Base:                github.String(base),
+				Body:                github.String(fOpt.PRDescription),
+				MaintainerCanModify: github.Bool(true),
+			})
+			if err != nil {
+				status[i].Err = err
+				continue
+			}
+		} // !dryRun
```
rickducott: Remove comment. Consider moving the loop to a helper function instead

pkgmgmtutils/update_formulas.go
```diff
@@ -0,0 +1,384 @@
...
+	if fOpt.WindowsShaRegex != "" {
+		byt = replaceSubmatch(byt, shas.windowsSha, regexp.MustCompile(fOpt.WindowsShaRegex))
+	}
+
+	return byt, nil
+}
+
+func updateAndPushLocalClone(version string, versionSha string, branchName string, commitString string, shas *sha256Outputs, fOpt *FormulaOptions, mergeRemote bool) error {
+	//
+	// Clone repo locally
```
rickducott: A lot of these section block comments don't add too much IMO

pkgmgmtutils/update_formulas.go
```diff
@@ -0,0 +1,384 @@
...
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"regexp"
+	"time"
+
+	"github.com/google/go-github/github"
+	"github.com/solo-io/go-utils/githubutils"
+	"github.com/solo-io/go-utils/versionutils"
+	"gopkg.in/src-d/go-git.v4"
```
rickducott: I like this library, nice find

scranton:
Any opinions on git commit names? These will appear in external repos (homebrew/homebrew-core and fishworks/fish-food), and if there are issues emails will be sent to bot@solo.io.

```
PRCommitName:    "Solo-io Bot",
PRCommitEmail:   "bot@solo.io",
```

Issue #124, opened 2019-04-16
port cobra cli utils from gloo


mitchdraft:
Looking to use https://github.com/solo-io/gloo/blob/master/projects/gloo/cli/pkg/testutils/utils.go in other projects (excluding the gloocli commands)

PR #134, opened 2019-04-21
extract clilog from gs

mitchdraft:
BOT NOTES: 
resolves https://github.com/solo-io/go-utils/issues/135
resolves https://github.com/solo-io/go-utils/issues/124
resolves https://github.com/solo-io/go-utils/issues/131


mitchdraft:
This will be updated with feedback from originating pr

clicore/logger.go
```diff
@@ -0,0 +1,165 @@
...
+		}
+		return nil
+	}
+
+	// if we decide we want to append logs, we can do it this way:
+	// (we would be required to first create the file and "rotate" it as it grows)
+	//file, err := os.OpenFile(path, os.O_WRONLY|os.O_APPEND, os.ModeAppend)
+
+	// for now, let's just create/overwrite the file each time
+	file, err := os.Create(path)
```
EItanya: fwiw this line will open a file in append mode if it exists, or create it if if doesn't

`file, err := os.OpenFile(path, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)`

clicore/logger.go
```diff
@@ -0,0 +1,165 @@
...
+
+	consoleStdoutCore := zapcore.NewCore(consoleEncoder, consoleInfo, stdOutMessages)
+	if verboseMode {
+		consoleStdoutCore = zapcore.NewCore(consoleEncoder, consoleInfo, stdOutMessagesVerbose)
+	}
+	consoleErrCore := zapcore.NewCore(consoleEncoder, consoleErrors, errorMessages)
+	return []zapcore.Core{consoleStdoutCore, consoleErrCore}
+}
+
+func soloPreferredEncoderConfig() zapcore.EncoderConfig {
```
EItanya: nit: `defaultEncoderConfig()`

Current one seems overly verbose

clicore/runner.go
```diff
@@ -0,0 +1,72 @@
...
+	"context"
+
+	"github.com/solo-io/go-utils/contextutils"
+	"github.com/spf13/cobra"
+	"go.uber.org/zap"
+)
+
+type CommandConfig struct {
+	Args                string
+	Command             func(context.Context, string) *cobra.Command
```
EItanya: nit: define a type for this function

mitchdraft: good idea, added https://github.com/solo-io/glooshot/pull/22

clicore/logger.go
mitchdraft: updated in https://github.com/solo-io/glooshot/pull/22

clicore/logger.go
mitchdraft: Interesting, I wonder if `O_APPEND` is a superset of `O_CREATE` because the line I referenced worked too. Either way, I think for now we don't want to risk growing massive log files so storing only the output from the last command seems right to me.

clicore/constants/constants.go
```diff
@@ -0,0 +1,5 @@
+package constants
```
EItanya: I'm wondering why this is in an entirely different package. I get that it is a constant but it's completely tied to the clicore main package and if I were a user of this package I would definitely look there first

mitchdraft: @EItanya there was an import cycle between clicore and contextutils

Issue #164, opened 2019-05-06
Make log level configurable


rickducott:
Currently defaults to standard production config level, which is Info.

PR #181, opened 2019-05-16
Set log level and improved support for Toml version parsing

rickducott:
BOT NOTES: 
resolves https://github.com/solo-io/go-utils/issues/180
resolves https://github.com/solo-io/go-utils/issues/164

### DIFF
--- a/versionutils/repo.go
+++ b/versionutils/repo.go
@@ -25,6 +25,12 @@ const (
 	SuperglooPkg = "github.com/solo-io/supergloo"
 )
 
+var (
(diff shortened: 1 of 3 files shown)


versionutils/repo.go
```diff
@@ -83,6 +105,7 @@ func ParseTomlFromDir(relativeDir string) ([]*toml.Tree, error) {
 	return parseTomlFromDir(relativeDir, constraint)
```
joekelley: let's deprecate this function and add ParseTomlConstraintsFromDir

rickducott: oops 👍

PR #190, opened 2019-05-21
Fix cache staleness

rickducott:
BOT NOTES: 
resolves https://github.com/solo-io/sm-marketplace/issues/452

### DIFF
--- a/installutils/kubeinstall/kube_installer.go
+++ b/installutils/kubeinstall/kube_installer.go
@@ -275,6 +275,12 @@ func (r *KubeInstaller) reconcileResources(ctx context.Context, installNamespace
 		return errors.Wrapf(err, "creating discovery rest mapper")
 	}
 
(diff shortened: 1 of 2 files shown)


installutils/kubeinstall/kube_installer.go
```diff
@@ -275,6 +275,12 @@ func (r *KubeInstaller) reconcileResources(ctx context.Context, installNamespace
 		return errors.Wrapf(err, "creating discovery rest mapper")
 	}
 
+	// refresh the client to get the new rest mappings for any crds created in the background (i.e. by a Job) since the client was last refreshed
```
rickducott: Some precedent for doing this already in L540. However, if a CRD was created in the background after the operator last ran "waitForResourceReady", it will have a stale cache that errors when trying to install resources for new CRDs. This manifests during istio install because the CRDs are created via a job, not directly as a resource during install.

PR #200, opened 2019-06-05
add github client getter that works with or without token

joekelley:
BOT NOTES: 
resolves https://github.com/solo-io/go-utils/issues/168


githubutils/repo.go
```diff
@@ -55,6 +55,21 @@ func GetClient(ctx context.Context) (*github.Client, error) {
 	return client, nil
 }
 
+func MustGetClient(ctx context.Context) *github.Client {
```
rickducott: Our use of `Must` typically means the function will either fatally exit or trigger and `Expect` failure as a potential side effect. Here, there isn't a side effect and we should probably use different terminology (both for consistency and to avoid training people that `Must` is safe, because other uses require care).

Issue #202, opened 2019-06-13
debugutils: create utility to gather all application data


EItanya:
gather all application data and logs into one place, then zip that information and save it.

PR #203, opened 2019-06-13
Debugutils

EItanya:
BOT NOTES: 
resolves https://github.com/solo-io/go-utils/issues/202


debugutils/aggregator.go
```diff
@@ -0,0 +1,129 @@
...
+func (lsc *LocalZipStorageClient) Read(file string) (io.ReadCloser, error) {
+	return lsc.fs.OpenFile(file, os.O_RDWR, 0777)
+}
+
+const (
+	aggregatorName = "aggregator"
+)
+
+type Aggregator struct {
+	col ResourceCollector
```
rickducott: What are we getting from all this shorthand? It just makes it confusing to try and remember what we're accessing below.

debugutils/logs.go
```diff
@@ -0,0 +1,152 @@
...
+type LogStorageClient struct {
+	fs  afero.Fs
+	dir string
+}
+
+func NewLogFileStorage(fs afero.Fs, dir string) *LogStorageClient {
+	return &LogStorageClient{fs: fs, dir: dir}
+}
+
+func (lfs *LogStorageClient) FetchLogs(requests []*LogsRequest) error {
```
rickducott: nit: suspect `lfs` was a pre-refactor artifact, not sure why we'd use it as shorthand.

debugutils/logs.go
```diff
@@ -0,0 +1,152 @@
...
+	eg := errgroup.Group{}
+	logsDir := filepath.Join(lfs.dir, "logs")
+	err := lfs.fs.Mkdir(logsDir, 0777)
+	if err != nil {
+		return err
+	}
+	lfs.dir = logsDir
```
rickducott: Handling of `lfs.dir` and `logsDir` is very strange. `logsDir` is derived from the previous value of `lfs.dir`. Generally overriding the struct value in this way to pass a new value through to where it is needed feels like an antipattern, should just be able to put it into a constructor (even if still making the directory lazily).

debugutils/logs.go
```diff
@@ -0,0 +1,152 @@
...
+	for _, request := range requests {
+		request := request
+		eg.Go(func() error {
+			reader, err := request.request.Stream()
```
rickducott: nit: the word "request" is too overloaded. Maybe call the inner request a `restRequest`, maybe leave a comment about why `request := request` is necessary / works.

debugutils/pods.go
```diff
@@ -0,0 +1,132 @@
...
+	client, err := kubernetes.NewForConfig(cfg)
+	if err != nil {
+		return nil, initializationError(err, labelPodFinderStr)
+	}
+	return &LabelPodFinder{
+		client: client,
+	}, nil
+}
+
+func (mpf *LabelPodFinder) GetPods(resources kuberesource.UnstructuredResources) ([]*corev1.PodList, error) {
```
rickducott: nit: why `mpf`

debugutils/resources.go
```diff
@@ -0,0 +1,213 @@
...
+		return nil, initializationError(err, resourceCollectorStr)
+	}
+	return &resourceCollector{
+		dynamicClient: dynamicClient,
+		restMapper:    restMapper,
+		podFinder:     podFinder,
+	}, nil
+}
+
+func (cc *resourceCollector) ResourcesFromManifest(manifests helmchart.Manifests, opts metav1.ListOptions) ([]kuberesource.VersionedResources, error) {
```
rickducott: nit: why `cc`

debugutils/resources.go
```diff
@@ -0,0 +1,213 @@
...
+	"k8s.io/client-go/dynamic"
+	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
+)
+
+const (
+	resourceCollectorStr = "resourceCollector"
+)
+
+var (
+	initializationError = func(err error, obj string) error {
```
rickducott: nit: since this is used across a bunch of abstractions probably makes sense to just go in an errors.go file

testutils/kube/supergloo_teardown.go
```diff
@@ -9,7 +9,7 @@ import (
 	. "github.com/onsi/gomega"
```
rickducott: Fix file name -- not specific to supergloo

debugutils/aggregator.go
```diff
@@ -0,0 +1,129 @@
...
+	pf  PodFinder
+	lsc *LogStorageClient
```
rickducott: I don't get why we need two different storage clients here - seems like a strange abstraction

debugutils/aggregator.go
```diff
@@ -0,0 +1,129 @@
...
+	"io"
+	"os"
+
+	"github.com/solo-io/go-utils/installutils/helmchart"
+	"github.com/solo-io/go-utils/tarutils"
+	"github.com/spf13/afero"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type ZipStorageClient interface {
```
rickducott: This isn't zip related

debugutils/aggregator.go
```diff
@@ -0,0 +1,129 @@
...
+	fs := afero.NewOsFs()
+	tmpd, err := afero.TempDir(fs, "", "")
+	if err != nil {
+		return nil, err
+	}
+	logStorage := NewLogFileStorage(fs, tmpd)
+	storageClient := NewLocalZipStorageClient(fs)
+	return &Aggregator{
+		pf:  podFinder,
+		col: collector,
```
rickducott: neither col nor collector make it clear that this is collecting kube resources

debugutils/resources.go
rickducott: nit: `RetrieveResourcesFromManifest`

debugutils/aggregator.go
rickducott: Actually, in addition to this there is storage logic in the ResourceCollector and LogRequestBuilder too -- I think we should revisit all the storage-related logic to simplify the abstraction and use a consistent abstraction across the ResourceCollector and log fetcher

debugutils/pods.go
```diff
@@ -0,0 +1,132 @@
...
+	corev1 "k8s.io/api/core/v1"
+	extensionsv1beta1 "k8s.io/api/extensions/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/kubernetes/pkg/apis/batch"
+)
+
+type PodFinder interface {
```
rickducott: Unit testing for this abstraction

debugutils/resources.go
```diff
@@ -0,0 +1,213 @@
...
+		return errors.Wrapf(err, "unable to initialize %s", obj)
+	}
+)
+
+type ResourceCollector interface {
```
rickducott: unit testing for this abstraction

debugutils/aggregator.go
```diff
@@ -0,0 +1,109 @@
...
+const (
+	aggregatorName = "aggregator"
+)
+
+type Aggregator struct {
+	resourceCollector ResourceCollector
+	logCollector      LogCollector
+	storageClient     StorageClient
+	fs                afero.Fs
+
```
rickducott: I would leave a comment here indicating it's the root directory for the collected files on `fs`

debugutils/aggregator.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+	unstructuredResources, err := manifest.ResourceList()
+	if err != nil {
+		return err
+	}
+	kubeResources, err := a.resourceCollector.RetrieveResources(unstructuredResources, namespace, metav1.ListOptions{})
+	if err != nil {
+		return err
+	}
+	if err := a.resourceCollector.SaveResources(filepath.Join(a.dir, "resources"), kubeResources); err != nil {
```
rickducott: pull "resources" and "logs" into consts, consider moving filepath.Join(...) into a private utility here.

debugutils/aggregator.go
```diff
@@ -0,0 +1,109 @@
...
+	dir string
+}
+
+func NewAggregator(resourceCollector ResourceCollector, logCollector LogCollector, storageClient StorageClient, fs afero.Fs, dir string) *Aggregator {
+	return &Aggregator{resourceCollector: resourceCollector, logCollector: logCollector, storageClient: storageClient, fs: fs, dir: dir}
+}
+
+func DefaultAggregator() (*Aggregator, error) {
+	fs := afero.NewOsFs()
+	storageClient := NewFileStorageClient(fs)
```
rickducott: Seems like we should be using the same storage client in the resource and log collector

debugutils/storage.go
```diff
@@ -0,0 +1,78 @@
...
+	client, err := storage.NewClient(ctx)
+	if err != nil {
+		return nil, err
+	}
+	return &GcsStorageClient{client: client, ctx: ctx}, nil
+}
+
+
+func (gsc *GcsStorageClient) Save(location string, resources ...*StorageObject) error {
+	bucket := gsc.client.Bucket("location")
```
rickducott: "location" -> location

debugutils/aggregator.go
```diff
@@ -0,0 +1,115 @@
...
+	resourceCollector, err := DefaultResourceCollector()
+	if err != nil {
+		return nil, errors.InitializationError(err, aggregatorName)
+	}
+	resourceCollector.storageClient = storageClient
```
rickducott: This is a bit hacky, it means we're using the wrong constructor. The aggregator requires the resource and log collector are implemented with the same underlying storage client (or at least, file system) or it won't work. To me this means:
1) We should never use the Default constructors from the aggregator, 2) We should change the collectors to take a storage client into the storage-related methods, or 3) change the collector interfaces to focus on collecting and split off the saving bits into other interfaces.

PR #204, opened 2019-06-18
Utilities for writing github bot applications and gcloud event subscribers

rickducott:
BOT NOTES: 
resolves https://github.com/solo-io/go-utils/issues/205


gcloudutils/util.go
```diff
@@ -0,0 +1,23 @@
...
+	"google.golang.org/api/cloudbuild/v1"
+)
+
+const (
+	nameConst    = "name"
+	versionConst = "version"
+	constraint   = "constraint"
+)
+
+func HandleFailedSourceBuild(ctx context.Context, build *cloudbuild.Build) error {
```
EItanya: No longer used as we don't have any more build which use the cloud `source` for the builds. Could be a good oppurtunity to trim that code

gcloudutils/tags.go
```diff
@@ -0,0 +1,77 @@
+package gcloudutils
```
EItanya: This file is missing yuval's latest changes involving installation ID

gcloudutils/clients.go
```diff
@@ -0,0 +1,20 @@
+package gcloudutils
+
+import (
+	"context"
+
+	"golang.org/x/oauth2/google"
+	"google.golang.org/api/cloudbuild/v1"
+)
+
+func NewCloudBuildClient(ctx context.Context) (*cloudbuild.Service, error) {
```
EItanya: This code doesn't allow for selecting any other set of credentials than the default, just want to make sure that is intended.

gcloudutils/clients.go
```diff
@@ -0,0 +1,20 @@
+package gcloudutils
```
EItanya: One of the nice things about this client set in solobot is the ability to get clients based on the project_id, would be nice to somehow port that logic here as well.

rickducott:
Looks like there's still a bunch of TODO's in that code, so I'll give it a few days.

gcloudutils/tags.go
rickducott: Cool, there are a bunch of changes missing and a bunch of TODOs -- I'll wait a bit and then update this.

rickducott:
Using this branch to stage a bunch of new utils pulled out of solobot to help write solo-io/post-release-bot and solo-io/changelog-bot

PR #211, opened 2019-06-21
Allow customizing the polling time in CurlEventuallyShouldRespond

yuval-k:


testutils/helper/curl.go
```diff
@@ -26,13 +26,21 @@ type CurlOpts struct {
...
+	currentTimeout := defaultTimeout
+	if len(timeout) > 0 {
+		currentTimeout = timeout[0]
+		if currentTimeout == 0 {
+			currentTimeout = defaultTimeout
+		}
+	}
+	pollingInterval := 5 * time.Second
+	if len(timeout) > 1 {
+		pollingInterval = timeout[1]
```
marcogschmidt: You could check for a zero value here as well, as with `currentTimeout` on line 34.

testutils/helper/curl.go
```diff
@@ -26,13 +26,21 @@ type CurlOpts struct {
 	WithoutStats bool
 }
 
-func (t *testContainer) CurlEventuallyShouldRespond(opts CurlOpts, substr string, ginkgoOffset int, timeout time.Duration) {
+func (t *testContainer) CurlEventuallyShouldRespond(opts CurlOpts, substr string, ginkgoOffset int, timeout ...time.Duration) {
 	defaultTimeout := time.Second * 20
-	if timeout == 0 {
-		timeout = defaultTimeout
+	currentTimeout := defaultTimeout
+	if len(timeout) > 0 {
```
marcogschmidt: Nit: maybe a `switch len(timeout)` where you overwrite the defaults in each `case` clause would be more readable.

testutils/helper/curl.go
yuval-k: not sure there is need to do so; the reason i do it the first line is to not break backwards compatibility,
but in the second one there is compatibility to break

marcogschmidt: Currently you can pass only exactly 1 timeout parameter, so no existing client code will pass the new `pollingInterval` parameter. I believe that as long as we use the original default in case `len(timeout) < 2` then it will be backwards compatible. anyway, I don't feel strongly about this, my concern was only about what might happen if you pass 0 as polling interval. I'm looking at the gomega code right now and it doesn't seem to protect against such a case.

yuval-k: yes, my intention is to remove the 0 from existing code.

PR #219, opened 2019-06-27
changelog reader abstraction

rickducott:
BOT NOTES: 
resolves https://github.com/solo-io/go-utils/issues/220


changelogutils/reader.go
```diff
@@ -0,0 +1,133 @@
+package changelogutils
+
+import (
+	"context"
+	"github.com/ghodss/yaml"
+	"github.com/pkg/errors"
```
EItanya: use go-utils errors, that way they can be directly compared with `errors.Is()`

PR #224, opened 2019-07-02
Config updates

rickducott:
BOT NOTES: 
resolves https://github.com/solo-io/go-utils/issues/223


osutils/client.go
```diff
@@ -0,0 +1,29 @@
...
+	"io/ioutil"
+	"os"
+)
+
+type OsClient interface {
+	Getenv(key string) string
+	ReadFile(path string) ([]byte, error)
+}
+
+type osClient struct {
```
EItanya: Not to add levels of abstraction, but how do you feel about adding an afero.fs here, or creating a seperate interface with afero. Creating utilities with it by default feels right to me.

botutils/botconfig/config.go
```diff
@@ -11,8 +11,34 @@ import (
...
+//go:generate mockgen -destination os_mock_test.go -self_package github.com/solo-io/go-utils/botutils/botconfig -package botconfig_test github.com/solo-io/go-utils/osutils OsClient
+
 const (
 	DefaultBotCfg  = "/etc/solo-github-app/config.yml"
+
+	BotConfigEnvVar = "BOT_CONFIG"
+	WebhookSecretEnvVar = "WEBHOOK_SECRET"
+	PrivateKeyEnvVar = "PRIVATE_KEY_FILE"
+	IntegrationIdEnvVar = "INTEGRATION_ID"
+	InstallationIdEnvVar = "INSTALLATION_ID"
```
EItanya: Installation ID should come from the request, not from the env. having it in the config was a temporary thing while we figured out how to pass it through the build object

botutils/botconfig/config.go
```diff
@@ -11,8 +11,34 @@ import (
...
+	FailedToParseEnvVarError = func(err error, name, value string) error {
+		return errors.Wrapf(err, "error parsing %s environment variable value: %s", name, value)
+	}
+
+	MissingBotConfigValueError = func(name string) error {
+		return errors.Errorf("missing important bot config, use %s to provide it or include in config map", name)
+	}
 )
 
 type Config struct {
```
EItanya: For such complex config with multiple sources of truth, might be worth looking at this https://github.com/spf13/viper
As I understand it, it's made to handle situations like this.

osutils/client.go
rickducott: discussed in person -- can implement with afero in the future, but this is used by bots who don't need the abstraction, and the interface is just for mock testing

botutils/botconfig/config.go
rickducott: Cool, going to remove the app config altogether -- it's bot specific. we will want to wire up build config at some point but don't have a solution for that right now

PR #226, opened 2019-07-04
tcp echo

EItanya:
BOT NOTES: 
resolves https://github.com/solo-io/go-utils/issues/227


testutils/helper/curl.go
```diff
@@ -26,11 +28,9 @@ type CurlOpts struct {
 	WithoutStats bool
 }
 
-func (t *testContainer) CurlEventuallyShouldRespond(opts CurlOpts, substr string, ginkgoOffset int, timeout ...time.Duration) {
+func (t *testContainer) getTimeouts(timeout ...time.Duration) (currentTimeout time.Duration, pollingInterval time.Duration) {
```
ilackarms: this method appears to have nothing to do with `testContainer`, want to remove the receiver? also how is this to be used?

EItanya: I will remove the reciever, also it was a part of the function which needed to be copied, so I put it all in one function. Yuval wrote it originally

testutils/kubernetes.go
```diff
@@ -104,6 +105,53 @@ func KubectlOutAsync(args ...string) (*bytes.Buffer, chan struct{}, error) {
 	return buf, done, err
 }
 
+func KubectlOutPipe(r io.Reader, args ...string) (<-chan *bytes.Buffer, chan struct{}, error) {
```
ilackarms: because of the name of this function, i expected to see the use of os pipes somewhere. it's not clear to me what this has to do with pipes. consider renaming the function?

i also see a lot of copy pasta in this file, maybe its time to refactor some of these Kubectl funcs?

Issue #230, opened 2019-07-09
Create a library for reasoning about kubernetes api versions


joekelley:
i.e. v1, v2alpha1, v2beta1

PR #231, opened 2019-07-09
kube api version util

joekelley:
BOT NOTES: 
resolves https://github.com/solo-io/go-utils/issues/230


kubeapiversion/api_version.go
```diff
@@ -0,0 +1,172 @@
...
+type PrereleaseModifier int
+
+var (
+	apiVersionRegexp *regexp.Regexp
+
+	MalformedVersionError = func(version string) error {
+		return errors.Errorf("Failed to parse kubernetes api version from %v", version)
+	}
+
+	InvalidMajorVersionError = errors.Errorf("Major version cannot be zero")
```
EItanya: if no formatting should be errors.New

kubeapiversion/api_version.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+
+	return &apiVersion{
+		major:      major,
+		prerelease: prerelease,
+		modifier:   modifier,
+	}, nil
+}
+
+func NewApiVersion(major, prerelease int, prereleaseModifier PrereleaseModifier) ApiVersion {
```
EItanya: Not sure this function should be public, it exposes the innards of the package which aren't necessary for consuming it

kubeapiversion/api_version.go
```diff
@@ -0,0 +1,172 @@
...
+
+func (v *apiVersion) Major() int {
+	return v.major
+}
+
+func (v *apiVersion) Prerelease() int {
+	return v.prerelease
+}
+
+func (v *apiVersion) PrereleaseModifier() PrereleaseModifier {
```
EItanya: I think PreReleaseModifier should probably have a `.String()` method to make it easier to consume

kubeapiversion/api_version_test.go
```diff
@@ -0,0 +1,134 @@
...
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/ginkgo/extensions/table"
+	. "github.com/onsi/gomega"
+	"github.com/solo-io/go-utils/kubeapiversion"
+)
+
+var _ = Describe("ApiVersion", func() {
+	Describe("ParseApiVersion", func() {
```
EItanya: Could you possibly also test some other error versions so it's a little more clear what versions are bad, also, maybe a link to kube documentation about how these versions work, and how our library differs from their sorting

Issue #243, opened 2019-07-18
Add usageutils to standardize interaction with the checkpoint server


rickducott:
Currently products inconsistently use check and report functions.

PR #244, opened 2019-07-18
Add usage client

rickducott:
BOT NOTES: 
resolves https://github.com/solo-io/go-utils/issues/243

### DIFF
--- a/usageutils/check.go
+++ b/usageutils/check.go
@@ -1,0 +1,35 @@
+package usageutils
+
+import (
+	check "github.com/solo-io/go-checkpoint"
+	"math/rand"
+	"time"
+)
+
+// A simple interface for interacting with the checkpoint server, for reporting and version checking
+type UsageClient interface {
+	Start(name, version string)
+}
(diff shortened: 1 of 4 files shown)


usageutils/check.go
```diff
@@ -0,0 +1,35 @@
...
+
+type usageClient struct {
+}
+
+func (c *usageClient) Start(name, version string) {
+	now := time.Now()
+	check.CallReport(name, version, now)
+	check.CallCheck(name, version, now)
+
+	jitter := time.Duration(rand.Intn(120)) * time.Minute
```
yuval-k: ideally jitter should be different in each iteration. maybe something like
```
for {
check.CallCheck
time.Sleep(time.Hour+time.Duration(rand.Intn(120)) * time.Minute)
}
```

see example here: https://github.com/hashicorp/consul/blob/53b25fbe383a42fa1cc80c9883bf0adaa9d79bb6/command/agent/agent.go#L120

PR #245, opened 2019-07-22
Get SHA256s from GitHub Release assets

scranton:


pkgmgmtutils/utils.go
```diff
@@ -22,6 +25,73 @@ type sha256Outputs struct {
...
+		case "linux":
+			shas.linuxSha, err = extractShaFromURL(f.GetBrowserDownloadURL())
+		case "windows":
+			shas.windowsSha, err = extractShaFromURL(f.GetBrowserDownloadURL())
+		}
+		if err != nil {
+			return nil, err
+		}
+	}
+	if shas.darwinSha == nil && shas.linuxSha == nil && shas.windowsSha == nil {
```
rickducott: I think this should be `||` since we want to error if any of them aren't found

scranton: I wasn't sure if we'd ever have cases where we'd only have 1 or 2 OSes provided, hence why the logic is testing for at least one. I can change logic and error message to reflect all three *must* be present if that's what we want.

PR #253, opened 2019-08-05
Build bots

rickducott:
BOT NOTES: 
resolves https://github.com/solo-io/go-utils/issues/252


gcloudutils/builders/mock.go
```diff
@@ -0,0 +1,54 @@
+package builders
```
EItanya: Is there a `go:generate` comamnd for these anywhere?
I don't see one.

gcloudutils/builders/repo_builder.go
```diff
@@ -0,0 +1,63 @@
+package builders
```
EItanya: This code can be deleted, we no longer build anything from source

PR #258, opened 2019-08-09
Kube dump fail handler

kdorosh:
related to: https://github.com/solo-io/gloo/issues/985

## Goal
Print kube pod logs and statuses in kube install namespace to ease debugging failed regression/e2e tests in CI

## Context
tested in gloo-ee PR and used to debug a failing regression test (printed the kube pod status with the failed image it was trying to pull, which was malformed)


kdorosh:
closing because I found similar logic in gloo

PR #268, opened 2019-08-29
Use lower case as is standard/recommended

kdorosh:
from the logrus [readme](https://github.com/sirupsen/logrus):

> **Seeing weird case-sensitive problems?** It's in the past been possible to import Logrus as both upper- and lower-case. Due to the Go package environment, this caused issues in the community and we needed a standard. Some environments experienced problems with the upper-case variant, so the lower-case was decided. Everything using logrus will need to use the lower-case: github.com/sirupsen/logrus. Any package that isn't, should be changed.

I was having numerous headaches trying to update the go-utils version in our rate limiting service, but after these changes (using https://github.com/solo-io/go-utils/pull/268/commits/44f8be641507b2ebac4fbb29738288d64fa54eed) I have no problem updating to the go-utils version in this commit hash. Oddly though, those changes prevent `go mod tidy` from working on go-utils itself -- requires more digging.

cc @EItanya @ilackarms 
BOT NOTES: 
resolves https://github.com/solo-io/go-utils/issues/269


kdorosh:
Other relevant docs: https://github.com/golang/go/wiki/Modules#how-can-i-resolve-parsing-gomod-unexpected-module-path-and-error-loading-module-requirements-errors-caused-by-a-mismatch-between-import-paths-vs-declared-module-identity

kdorosh:
Ah, I think the current behavior is desired, although annoying, until helm>k8s>docker updates its dependency on capitalized Sirupsen

PR #271, opened 2019-08-30
Parse image from toml

mitchdraft:
BOT NOTES: 
resolves https://github.com/solo-io/go-utils/pull/279

### DIFF
--- a/versionutils/repo.go
+++ b/versionutils/repo.go
@@ -207,8 +207,8 @@ func NewSimpleTomlParserFromString(content string) (*SimpleTomlParser, error) {
 // name = "name-2"
 // other = "some other value 2"
 //
-// GetTomlValues("my-file.toml", "rootTableName", "name", "name-1")
-// would return map[string]string{"name":"name-1","other":"some other value"}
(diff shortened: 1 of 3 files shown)


versionutils/repo.go
```diff
@@ -135,6 +136,7 @@ func parseTomlFromDir(relativeDir, configType string) ([]*toml.Tree, error) {
 type TomlWrapper struct {
 	Overrides   []*toml.Tree
 	Constraints []*toml.Tree
+	SoloIo      []*toml.Tree
```
mitchdraft: don't need this with the multiple toml file approach
(had initially thought to overload the `Gopkg.toml` file)

mitchdraft:
Since dep warns when unrecognized keys are present in the Gopkg.toml file, it's not possible to add image versions there. We would need to manage a separate image.toml file, which is what this util was designed to parse.
After speaking with @ilackarms, we questioned whether toml is a filetype that we would wish to do more with (no, since parsing yaml is easier) - and even whether we need a designated image version tracking system (probably not)

closing in favor of just specifying the target image as a constant in a go pkg

Issue #281, opened 2019-09-11
use BeEquivalentTo instead of Equal in ExpectEqualProtoMessages


joekelley:
replaces 
```
    Expected
        <string>: "applic..."
    to equal       |
        <string>: "mesh:<..."
```

with
```
    Expected
        <string>: application_state:<installation_state:<install_namespace:"linkerd-ns" > status:<> metadata:<> registry_name:"hub-meshes" application_name:"linkerd2" > 
    to be equivalent to
        <string>: mesh:<status:<> metadata:<name:"istio-mesh" namespace:"test-install" resource_version:"2" > istio:<installation_namespace:"istio-system" version:"1.0.6" > > 
```

😤

PR #282, opened 2019-09-11
Print bigger diff on ExpectEqualProtoMessages failure

joekelley:
BOT NOTES: 
resolves https://github.com/solo-io/go-utils/issues/281


testutils/assertions.go
```diff
@@ -6,23 +6,21 @@ import (
...
 	if proto.Equal(a, b) {
 		return
 	}
-	// One shortcoming is that you only get +/- 5 chars of context
-	// per: https://github.com/onsi/gomega/blob/master/format/format.go#L146
-	// TODO(mitchdraft) gomega pr to modify charactersAroundMismatchToInclude (if not merged will make a util)
-	Expect(a.String()).To(Equal(b.String()), optionalDescription...)
+	// Prefer BeEquivalentTo over Equal because it prints entire strings for
+	// the actual and expected values.
+	Expect(a.String()).To(BeEquivalentTo(b.String()), optionalDescription...)
```
mitchdraft: now that the gomega pr merged you need to bump gomega to v1.7.0 and set `CharactersAroundMismatchToInclude` to a large number
might want to do
```
initalCharactersToInclude := CharactersAroundMismatchToInclude
CharactersAroundMismatchToInclude = 10000
```
before the `Expect`, and:
`CharactersAroundMismatchToInclude = initialCharactersToInclude`
after the `Expect`

mitchdraft: oh I see what you are doing, how about this:
```
if a.String() != b.String() {
  fmt.Fprintf("protos do not match\nhave (length %v):\n%v\n want (length %v):\n %v", len(a.String()),a.String(),len(b.String()),b.String())
  fmt.Fprintf(GinkgoWriter, "set CharactersAroundMismatchToInclude to include more or less diff context (current value is )", CharactersAroundMismatchToInclude)
}
Expect(a.String()).To(Equal(b.String()), optionalDescription...)
```

i.e. use both approaches

may as well be generous with the clues on this central assertion util

testutils/assertions.go
```diff
@@ -19,10 +23,9 @@ import (
 //       <string>: "...-10101" ver..."
 func ExpectEqualProtoMessages(a, b proto.Message, optionalDescription ...interface{}) {
 	if proto.Equal(a, b) {
```
mitchdraft: I think I mislead you with the structure of my pseudocode
function should return with no output if `proto.Equal(a, b)`

testutils/assertions.go
```diff
@@ -19,10 +23,9 @@ import (
...
-		return
+		fmt.Fprintf(ginkgo.GinkgoWriter, "protos do not match\nhave (length %v):\n%v\n want (length %v):\n %v", len(a.String()), a.String(), len(b.String()), b.String())
```
mitchdraft: nit: want to use backtick multiline strings or do something like this:
```
fmt.Fprintf(ginkgo.GinkgoWriter, "protos do not match\n" +
"have (length %v):\n" +
"%v\n" +
"want (length %v):\n"+
"%v",
len(a.String()), a.String(), len(b.String()), b.String())
```

joekelley:
Closing as stale. I still like this idea in general, but there has been some resistance to printing huge strings when it's not obvious what the diff is.

PR #286, opened 2019-09-17
Add test util method

GrahamGoudeau:


manifesttestutils/util.go
```diff
@@ -27,6 +27,7 @@ import (
 type TestManifest interface {
 	ExpectDeployment(deployment *v1beta1.Deployment) *v1beta1.Deployment
 	ExpectDeploymentAppsV1(deployment *appsv1.Deployment)
+	ExpectNoSuchResource(kind, namespace, name string)
```
joekelley: I like this idea, but wondering if you could instead use `Expect(...).To(BeNil())` or similar instead

joekelley: https://godoc.org/github.com/onsi/gomega#Assertion

Issue #291, opened 2019-09-19
Improve TOML parsing


marcogschmidt:
Currently the .toml parsing utilities only consider the `version` attribute on `constraints`/`overrides`. We need to update it to also look for `revision` and `branch`.

PR #292, opened 2019-09-19
Get dependency version from revision/branch fields

marcogschmidt:
Consider all possible version fields (`version`, `branch`, `revision`) when retrieving a dependency version from a .toml file.
BOT NOTES: 
resolves https://github.com/solo-io/go-utils/issues/291

### DIFF
--- a/versionutils/repo.go
+++ b/versionutils/repo.go
@@ -164,7 +164,7 @@ func getVersionFromTree(tomlTree *toml.Tree, pkgName string) (version string, fo
 	}
 
 	if tomlTree.Get(nameConst) != pkgName {
-		return
+		return "", false
 	}
 
 	switch {
@@ -175,5 +175,5 @@ func getVersionFromTree(tomlTree *toml.Tree, pkgName string) (version string, fo
 	case !isEmpty(tomlTree, branchConst):
 		return tomlTree.Get(branchConst).(string), true
 	}
-	return
+	return "", false
 }
--- a/versionutils/repo.go
+++ b/versionutils/repo.go
@@ -14,11 +14,13 @@ import (
 )
 
 const (
-	gopkgToml    = "Gopkg.toml"
-	constraint   = "constraint"
-	override     = "override"
-	nameConst    = "name"
-	versionConst = "version"
+	gopkgToml     = "Gopkg.toml"
+	constraint    = "constraint"
+	override      = "override"
+	nameConst     = "name"
+	versionConst  = "version"
+	revisionConst = "revision"
+	branchConst   = "branch"
 
 	GlooPkg      = "github.com/solo-io/gloo"
 	SoloKitPkg   = "github.com/solo-io/solo-kit"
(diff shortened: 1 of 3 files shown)


versionutils/repo.go
```diff
@@ -155,3 +157,19 @@ func ParseFullTomlFromDir(relativeDir string) (*TomlWrapper, error) {
 func ParseFullToml() (*TomlWrapper, error) {
 	return ParseFullTomlFromDir("")
 }
+
+func getVersionFromTree(tomlTree *toml.Tree, pkgName string) (found bool, version string) {
```
kdorosh: feels more idiomatic to reverse the order of the return args

kdorosh: e.g. checking if a key is valid in a golang map:
```golang
if val, ok := map[key]; ok {
  ...
}
```

marcogschmidt: Right! Will update

versionutils/repo.go
```diff
@@ -155,3 +157,19 @@ func ParseFullTomlFromDir(relativeDir string) (*TomlWrapper, error) {
...
+	isEmpty := func(node *toml.Tree, key string) bool {
+		return node.Get(key) == nil || node.Get(key) == ""
+	}
+
+	switch {
+	case tomlTree.Get(nameConst) == pkgName && !isEmpty(tomlTree, versionConst):
+		return true, tomlTree.Get(versionConst).(string)
+	case tomlTree.Get(nameConst) == pkgName && !isEmpty(tomlTree, revisionConst):
+		return true, tomlTree.Get(revisionConst).(string)
+	case tomlTree.Get(nameConst) == pkgName && !isEmpty(tomlTree, branchConst):
```
kdorosh: can we stop repeating `tomlTree.Get(nameConst) == pkgName` in the switch statement by checking this before the switch statement?

versionutils/repo.go
```diff
@@ -155,3 +157,23 @@ func ParseFullTomlFromDir(relativeDir string) (*TomlWrapper, error) {
...
 	return ParseFullTomlFromDir("")
 }
+
+func getVersionFromTree(tomlTree *toml.Tree, pkgName string) (version string, found bool) {
+	isEmpty := func(node *toml.Tree, key string) bool {
+		return node.Get(key) == nil || node.Get(key) == ""
+	}
+
+	if tomlTree.Get(nameConst) != pkgName {
+		return
```
kdorosh: the naked return is only supposed to be used in super short functions (~3-5 lines of code)

i think we should explicitly return `"", false` here

versionutils/repo.go
```diff
@@ -155,3 +157,23 @@ func ParseFullTomlFromDir(relativeDir string) (*TomlWrapper, error) {
...
+
+	switch {
+	case !isEmpty(tomlTree, versionConst):
+		return tomlTree.Get(versionConst).(string), true
+	case !isEmpty(tomlTree, revisionConst):
+		return tomlTree.Get(revisionConst).(string), true
+	case !isEmpty(tomlTree, branchConst):
+		return tomlTree.Get(branchConst).(string), true
+	}
+	return
```
kdorosh: the naked return is only supposed to be used in super short functions (~3-5 lines of code)

i think we should explicitly return `"", false` here

PR #293, opened 2019-09-19
Return more info about dep & git versions

marcogschmidt:
Improve the commands to get version info from the `Gopkg.toml` file and from a local git repo.


changelog/v0.10.8/more-version-info.yaml
```diff
@@ -0,0 +1,5 @@
+changelog:
+- type: NON_USER_FACING
+  description: Return more info about a dependency version detrieved from a .toml file.
```
EItanya: retrieved/derived?

versionutils/git/version_test.go
```diff
@@ -0,0 +1,21 @@
...
+
+	It("works as expected", func() {
+		refInfo, err := git.GetGitRefInfo("./../..")
+		Expect(err).NotTo(HaveOccurred())
+		Expect(refInfo).NotTo(BeNil())
+		// Git commit sha is always 40 characters long
+		Expect(refInfo.Hash).To(HaveLen(40))
+		// Our tags always start with"v"
+		Expect(refInfo.Tag).To(HavePrefix("v"))
+		// Nothing reliable to test the branch with, unfortunately
```
EItanya: I kind of just want to make sure it doesn't crash because you are array explicitly indexing, does that make sense?

versionutils/git/version.go
```diff
@@ -0,0 +1,122 @@
...
+
+func (g gitRepo) getBranch() (string, error) {
+	cmd := exec.Command("git", "branch")
+	cmd.Dir = g.relativeDir
+	output, err := cmd.Output()
+	if err != nil {
+		return "", FailedCommandError(err, cmd.Args, string(output))
+	}
+
+	lines := strings.Split(string(output), "\n")
```
EItanya: `bufio.Scanner`
Might be overkill for this, but best solution for reading lines

marcogschmidt: Nice, never used it!

versionutils/git/version_test.go
marcogschmidt: Removed this test, but added a check for the array length

PR #334, opened 2019-10-29
Refresh operator cache

ashleywang1:


changelog/v0.10.24/refresh-operator-cache.yaml
```diff
@@ -0,0 +1,5 @@
+changelog:
+  - type: FIX
+    description: Allow updates to a KubeInstaller cache
+    issueLink: https://github.com/solo-io/sm-marketplace/issues/802
```
rickducott: should link to an issue from a public repo so the link will be accessible to non-solo people

installutils/kubeinstall/kube_installer.go
```diff
@@ -613,6 +614,10 @@ func (r *KubeInstaller) ListAllResources(ctx context.Context) kuberesource.Unstr
 	return r.cache.List()
 }
 
+func (r *KubeInstaller) RefreshResourcesCache(cache *Cache) {
```
rickducott: not totally sure how this is intended to be used -- if updated in the background from a goroutine, maybe we should protect this with a mutex? if updated prior to using the installer, then why not just create a new installer?

rickducott: or by re-calling "Init" on the cache object? 

Not really opposed to this change, just not sure i can see the full picture / can't reason about the tradeoffs of this vs. alternative approach

PR #335, opened 2019-10-30
Add Refresh function to a kubeinstall.Cache

ashleywang1:


installutils/kubeinstall/cache.go
```diff
@@ -49,6 +49,15 @@ func (c *Cache) Init(ctx context.Context, cfg *rest.Config, filterFuncs ...kuber
 	return nil
 }
 
+/*
+Refresh the cache with the snapshot of the current cluster
+*/
+func (c *Cache) Refresh(ctx context.Context, cfg *rest.Config, filterFuncs ...kuberesource.FilterResource) error {
+	// unlock cache after sync is complete
+	c.access.Lock()
+	return c.Init(ctx, cfg, filterFuncs...)
```
joekelley: We could extract the body of the Init function into a new function that returns cluster resources, then we can run that function without locking access to the cache, and only lock when we swap out the contents of the cache.

As is, if we were to refresh the cache every 60 seconds, it might only be available for 15-30 seconds at a time.

PR #339, opened 2019-11-07
Update logrus take 1

rickducott:
After this, I can pull in go-utils as long as I use this replace in my downstream projects: 
`replace github.com/Sirupsen/logrus => github.com/sirupsen/logrus v1.0.6`

And I also make sure I don't pull in a package of Sirupsen as an indirect dependency. 
BOT NOTES: 
resolves https://github.com/solo-io/go-utils/issues/340


rickducott:
If it passes CI I'll add a CL

PR #349, opened 2019-11-21
Upgrade Helm libraries to 3.0.0

GrahamGoudeau:


GrahamGoudeau:
/kick https://console.cloud.google.com/cloud-build/builds/98dd88e0-8d7e-43df-9780-787a2efb57d0?project=solo-public
timed out waiting for tcp echo pod to start up

certutils/gen_cert.go
```diff
@@ -6,8 +6,9 @@ import (
 	"crypto/rand"
 	"crypto/rsa"
 
+	"k8s.io/kubernetes/staging/src/k8s.io/client-go/util/cert"
```
EItanya: This isn't correct. The staging libraries are transferred to their own repos and then consumed from there.

PR #355, opened 2019-12-14
updated hashutils

EItanya:
BOT NOTES: 
resolves https://github.com/solo-io/go-utils/issues/356


cloudbuild.yaml
```diff
@@ -28,7 +28,7 @@ steps:
     waitFor: ['build']
     id: 'run-tests'
 
-  - name: 'gcr.io/$PROJECT_ID/go-mod-ginkgo:0.1.5'
+  - name: 'gcr.io/$PROJECT_ID/go-mod-ginkgo.slim'
```
rickducott: Why is this not tagged with a version?

EItanya: Accident, going to remove

hashutils/hash.go
```diff
@@ -30,3 +35,39 @@ func hashValue(val interface{}) uint64 {
 	}
 	return h
 }
+
+type SafeHasher interface {
+	Hash(hasher hash.Hash64) (uint64, error)
```
yuval-k: comment explaining the semantics - is this hasher reset? or just appended to?

hashutils/hash.go
```diff
@@ -30,3 +35,39 @@ func hashValue(val interface{}) uint64 {
...
+	return hasher.Sum64(), nil
+}
+
+func hashValueSafe(hasher hash.Hash64, val interface{}) error {
+	if hashObj, ok := val.(SafeHasher); ok {
+		_, err := hashObj.Hash(hasher)
+		if err != nil {
+			return err
+		}
+		return nil
```
yuval-k: nit: instead of 
```

		if err != nil {
			return err
		}
		return nil 
```
you can do
```
return err
```

hashutils/hash.go
```diff
@@ -30,3 +35,39 @@ func hashValue(val interface{}) uint64 {
...
+	}
+	h, err := hashstructure.Hash(val, nil)
+	if err != nil {
+		return err
+	}
+	if err = binary.Write(hasher, binary.LittleEndian, h); err != nil {
+		return err
+	}
+	return nil
```
yuval-k: nit: 
```
return binary.Write(hasher, binary.LittleEndian, h)
```

PR #358, opened 2019-12-17
Changelog release

rickducott:
BOT NOTES: 
resolves https://github.com/solo-io/go-utils/issues/357


changelogutils/validator.go
```diff
@@ -15,10 +18,18 @@ import (
 //go:generate mockgen -destination repo_client_mock_test.go -self_package github.com/solo-io/go-utils/changelogutils -package changelogutils_test github.com/solo-io/go-utils/githubutils RepoClient
 
 const (
-	MasterBranch = "master"
+	MasterBranch           = "master"
+	ValidationSettingsFile = "validation.yaml"
```
EItanya: Kind of a nit: but I'm wondering if we can include this info in our already created solo-project.yaml
Adding another yaml file for one setting seems like overkill when we already have a CI config file for another purpose.
Granted these have seperate uses right now, but this might be a good oppurtunity to combine them

rickducott: The solo-project.yaml is not an option because go-utils can't depend on that repo - the dependency is in the opposite direction. I'm open to refactoring later, but Gloo is blocked now and the config needs to live somewhere.

Issue #363, opened 2019-12-19
DRY out stats server startup config


GrahamGoudeau:
Almost every docker image we ship has an explicit check on START_STATS_SERVER. There's no reason this couldn't be DRY'd out.

PR #364, opened 2019-12-19
DRY out stats server startup config

GrahamGoudeau:
BOT NOTES: 
resolves https://github.com/solo-io/go-utils/issues/363


changelog/v0.11.5/dry-stats.yaml
```diff
@@ -0,0 +1,4 @@
+changelog:
+  - type: FIX
```
kdorosh: non user facing

kdorosh: eh since this is go-utils this is fine as a fix too

Issue #367, opened 2019-12-20
hashutils - must hash


EItanya:
The safe hash function now takes in a hasher, and returns an error. Which can be confusing/tedious to call. Create a utility which handles this complexity for the user.

PR #368, opened 2019-12-20
MustHash

EItanya:
BOT NOTES: 
resolves https://github.com/solo-io/go-utils/issues/367

### DIFF
--- a/hashutils/hash.go
+++ b/hashutils/hash.go
@@ -99,7 +99,7 @@ func HashableEqual(val1 interface{}, val2 interface{}) (equal bool, ok bool) {
 }
 
 /*
-	This function will take hash anything, and panic if it fails.
+	This function will hash anything, and panic if it fails.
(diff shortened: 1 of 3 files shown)


hashutils/hash.go
```diff
@@ -96,3 +97,21 @@ func HashableEqual(val1 interface{}, val2 interface{}) (equal bool, ok bool) {
 	}
 	return false, false
 }
+
+func MustHash(val interface{}) uint64 {
```
GrahamGoudeau: Can we document this function's behavior in a comment?

hashutils/hash.go
```diff
@@ -96,3 +97,26 @@ func HashableEqual(val1 interface{}, val2 interface{}) (equal bool, ok bool) {
 	}
 	return false, false
 }
+
+/*
+	This function will take hash anything, and panic if it fails.
```
GrahamGoudeau:
```suggestion
	This function will hash anything, and panic if it fails.
```

PR #386, opened 2020-02-01
add WithExistingLogger

yuval-k:
this will allow us to see debug log in ext auth
BOT NOTES: 
resolves https://github.com/solo-io/go-utils/issues/387

### DIFF
--- a/loggingutils/grpc.go
+++ b/loggingutils/grpc.go
@@ -7,7 +7,7 @@ import (
 	"google.golang.org/grpc"
 )
 
-func LoggerInterceptor(logger *zap.SugaredLogger) grpc.UnaryServerInterceptor {
+func GrpcUnaryServerLoggerInterceptor(logger *zap.SugaredLogger) grpc.UnaryServerInterceptor {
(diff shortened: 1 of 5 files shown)


loggingutils/grpc.go
```diff
@@ -0,0 +1,15 @@
+package loggingutils
+
+import (
+	"context"
+	"github.com/solo-io/go-utils/contextutils"
+	"go.uber.org/zap"
+	"google.golang.org/grpc"
+)
+
+func LoggerInterceptor(logger *zap.SugaredLogger) grpc.UnaryServerInterceptor {
```
marcogschmidt: Since there are other types of interceptors we might want to implement in the future (`UnaryClientInterceptor`, `StreamServerInterceptor`, `StreamClientInterceptor`), can we use a more specific name for this constructor? Maybe `GrpcUnaryServerLoggerInterceptor` (even though that's a mouthful...)?

yuval-k: good point

PR #391, opened 2020-02-11
fix logging timestamps if stats server is enabled

ashleywang1:
In StartStatsServerWithPort, we set the fallback logger for a ctx with `contextutils.SetFallbackLogger(logger.Sugar())`, building a default logger that uses an epoch timestamp.

By default, we actually want the zapcore.ISO8601TimeEncoder format instead since it is more human-readable.

### DIFF
--- a/stats/stats.go
+++ b/stats/stats.go
@@ -2,6 +2,7 @@ package stats
 
 import (
 	"fmt"
+	"go.uber.org/zap/zapcore"
 	"os"
 	"sort"
 	"text/template"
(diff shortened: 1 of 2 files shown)


stats/stats.go
```diff
@@ -56,6 +57,7 @@ func StartStatsServerWithPort(startupOpts StartupOptions, addhandlers ...func(mu
 	}
 
 	logconfig := zap.NewProductionConfig()
+	logconfig.EncoderConfig.EncodeTime = zapcore.ISO8601TimeEncoder
```
GrahamGoudeau: Do we want to instead just pass in a logger to use? That'll ensure it's always consistent with whoever uses this (someone may not be using the ISO8601TimeEncoder)

ilackarms:
Can we also use this pr as an opportunity to do debug logging at boot time?

Issue #392, opened 2020-02-14
Configure Helm Installer with kube context


harveyxia:
Prior to installation, the Helm installer checks for the existence of the specified namespace for installation. However, its namespace client [assumes the current context](https://github.com/solo-io/go-utils/blob/9505ac8feb5377eaf24b09c64ce4c45b606a12dc/installutils/helminstall/client.go#L193) for this check, an assumption that doesn't hold in multicluster contexts.

PR #393, opened 2020-02-14
Add kube context to helm installer

harveyxia:
BOT NOTES: 
resolves https://github.com/solo-io/go-utils/issues/392

### DIFF
--- a/installutils/helminstall/client.go
+++ b/installutils/helminstall/client.go
@@ -23,19 +23,19 @@ const (
 // This interface implements the Helm CLI actions. The implementation relies on the Helm 3 libraries.
 type HelmClient interface {
 	// Prepare an installation object that can then be .Run() with a chart object
-	NewInstall(kubeContext, namespace, releaseName string, dryRun bool) (HelmInstaller, *cli.EnvSettings, error)
+	NewInstall(kubeConfig, kubeContext, namespace, releaseName string, dryRun bool) (HelmInstaller, *cli.EnvSettings, error)
 
 	// Prepare an un-installation object that can then be .Run() with a release name
-	NewUninstall(helmKubeContext, namespace string) (HelmUninstaller, error)
+	NewUninstall(kubeConfig, kubeContext, namespace string) (HelmUninstaller, error)
 
 	// List the already-existing releases in the given namespace
-	ReleaseList(helmKubeContext, namespace string) (ReleaseListRunner, error)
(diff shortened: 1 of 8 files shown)


installutils/helminstall/client.go
```diff
@@ -17,24 +17,25 @@ const (
...
 	TempChartPrefix          = "temp-helm-chart"
 	helmNamespaceEnvVar      = "HELM_NAMESPACE"
+	helmKubeContextEnvVar    = "HELM_KUBECONTEXT"
 )
 
 // This interface implements the Helm CLI actions. The implementation relies on the Helm 3 libraries.
 type HelmClient interface {
 	// Prepare an installation object that can then be .Run() with a chart object
-	NewInstall(namespace, releaseName string, dryRun bool) (HelmInstaller, *cli.EnvSettings, error)
+	NewInstall(kubeConfig, kubeContext, namespace, releaseName string, dryRun bool) (HelmInstaller, *cli.EnvSettings, error)
```
GrahamGoudeau: I guess we could also stick these two new args on the constructor, and eliminate having to repeat them on every method invocation

installutils/helminstall/installer.go
```diff
@@ -35,6 +35,9 @@ type Installer interface {
 }
 
 type InstallerConfig struct {
+	KubeConfig string
```
GrahamGoudeau: super minor thing, but let's make sure that both fields get some brief documentation

installutils/helminstall/client.go
```diff
@@ -190,7 +191,7 @@ func (d *defaultHelmClient) ReleaseExists(namespace, releaseName string) (bool,
 
 // Build a Helm EnvSettings struct
 // basically, abstracted cli.New() into our own function call because of the weirdness described in the big comment below
-func NewCLISettings(namespace string) *cli.EnvSettings {
+func NewCLISettings(kubeConfig, kubeContext, namespace string) *cli.EnvSettings {
 	// The installation namespace is expressed as a "config override" in the Helm internals
 	// It's normally set by the --namespace flag when invoking the Helm binary, which ends up
```
GrahamGoudeau: can we generalize this comment a bit? Because now it applies to more than just the one value that comes in over an env var

installutils/helminstall/client.go
harveyxia: Let's leave it in the methods for now so that the same instance of HelmClient can be used to install across cluster/namespace, preventing the need for the caller to make a HelmClientFactory

PR #405, opened 2020-05-05
extend HelmClient to accept in-memory kubeconfig

harveyxia:
BOT NOTES: 
resolves https://github.com/solo-io/go-utils/issues/406


harveyxia:
Haven't written tests yet, wanted to get feedback on the approach first

installutils/helminstall/internal/memory_backed_rest_client_getter.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+	return ToRESTMapper(cfg)
+}
+
+func (i *inMemoryRESTClientGetter) ToRawKubeConfigLoader() clientcmd.ClientConfig {
+	return i.rawConfig
+}
+
+//
+// files below this comment were shamelessly and brazenly stolen from the Helm codebase
```
joekelley, line 64: if so, could be helpful to link to the source

installutils/helminstall/internal/memory_backed_rest_client_getter.go
```diff
@@ -0,0 +1,90 @@
...
+var (
+	defaultCacheDir = filepath.Join(clientcmd.RecommendedConfigDir, "http-cache")
+
+	// overlyCautiousIllegalFileCharacters matches characters that *might* not be supported.  Windows is really restrictive, so this is really restrictive
+	overlyCautiousIllegalFileCharacters = regexp.MustCompile(`[^(\w/\.)]`)
+
+	removeProtocol = regexp.MustCompile("http[s]?:/{2}")
+)
+
+func NewInMemoryRESTClientGetter(rawConfig clientcmd.ClientConfig) genericclioptions.RESTClientGetter {
```
GrahamGoudeau, line 27: right, so I guess we only need this for helm stuff, so it's fine to stash it away in `internal/` here?

installutils/helminstall/types/interfaces.go
```diff
@@ -1,30 +1,45 @@
...
 	// Prepare an un-installation object that can then be .Run() with a release name
-	NewUninstall(kubeConfig, kubeContext, namespace string) (HelmUninstaller, error)
+	NewUninstall(namespace string) (HelmUninstaller, error)
 
 	// List the already-existing releases in the given namespace
-	ReleaseList(kubeConfig, kubeContext, namespace string) (ReleaseListRunner, error)
+	ReleaseList(namespace string) (ReleaseListRunner, error)
 
 	// Returns the Helm chart archive located at the given URI (can be either an http(s) address or a file path)
 	DownloadChart(chartArchiveUri string) (*chart.Chart, error)
```
GrahamGoudeau, line 32: probably worth noting that this method will still use the filesystem, and will fail if your filesystem is readonly (so worth checking whether we depend on this in the csr agent installer)

installutils/helminstall/helm_client.go
```diff
@@ -6,11 +6,18 @@ import (
...
+func DefaultHelmClientFileConfigFactory() types.HelmClientForFileConfigFactory {
+	return internal.NewHelmClientForFileConfigFactory(
+		internal.NewFs(afero.NewOsFs()),
+		internal.NewDefaultResourceFetcher(),
+		internal.NewHelmFactories())
+}
+
+// HelmClient factory that accepts kubeconfig in memory.
+func DefaultHelmClientMemoryConfigFactory() types.HelmClientForMemoryConfigFactory {
+	return internal.NewHelmClientForMemoryConfigFactory(
```
joekelley, line 19: The approach in this PR seems solid overall, my only question is whether we want to embrace the factory pattern to this extent within this go-utils package. Here we have a provider for a factory that returns a helm client. Users might be happy with just `NewHelmClientForMemoryConfig`, but it's in `internal`. My preference would be to have go-utils expose simple constructors and leave the factory type definitions and implementations to importing projects, wdyt?

installutils/helminstall/internal/memory_backed_rest_client_getter.go
harveyxia: I was confused by this as well, the existing code seemed to hide the HelmClient away in `internal` but exposed default constructors through in "installutils/helminstall/helm_client.go", so I adopted the same pattern here. Happy to change it to something else though

installutils/helminstall/internal/memory_backed_rest_client_getter.go
harveyxia: @GrahamGoudeau do you recall where in the Helm codebase you copy pasta-ed this from?

installutils/helminstall/helm_client.go
harveyxia: Yup, that makes sense to me! I've replaced the factory constructors here with regular HelmClient constructors that expose the relevant in-mem vs. file config params

installutils/helminstall/installer.go
```diff
@@ -37,16 +33,14 @@ type installer struct {
 	out          io.Writer
 }
 
-func MustInstaller() types.Installer {
-	cfg, err := kubeutils.GetConfig("", "")
-	if err != nil {
-		log.Fatal(err)
+type InstallerFactory func(helmClient types.HelmClient) types.Installer
```
joekelley, line 36: could we do away with this factory as well per the discussion in a previous comment, and simply expose `NewInstaller`?

installutils/helminstall/helm_client.go
```diff
@@ -4,13 +4,26 @@ import (
...
+		internal.NewHelmFactories(),
+		kubeConfig,
+		kubeContext,
+	)
+}
+
+// HelmClient factory that accepts kubeconfig in memory.
+func DefaultHelmClientMemoryConfig(config clientcmd.ClientConfig) types.HelmClient {
+	return internal.NewHelmClientForMemoryConfig(
+		internal.NewFs(afero.NewOsFs()),
```
joekelley, line 24: should the afero filesystem implementation be injectable by the user for cases where we don't have access to read/write on the host file system?

installutils/helminstall/installer.go
harveyxia: ah missed this one

PR #408, opened 2020-06-23
Stop opening PRs to homebrew-core for non-stable releases

GrahamGoudeau:
BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/3178


go.mod
```diff
@@ -44,7 +44,6 @@ require (
 	goji.io v2.0.2+incompatible
 	golang.org/x/oauth2 v0.0.0-20190604053449-0f29369cfe45
 	golang.org/x/sync v0.0.0-20190911185100-cd5d95a43a6e
-	golang.org/x/xerrors v0.0.0-20190717185122-a985d3407aa7
```
mlholland, line 47: Quick Question: What are go.mod and go.sum and why do you need to change these when the real changes for this PR seem to just be logic changes in update_formulas.go?

kdorosh: go.mod defines the dependencies for the project, go.sum is a lock file that fully enumerates all the dependencies required to build the module

if import statements change or pull in new versions in go code, it is expected to have go.mod and go.sum change

kdorosh: as for why they changed in this PR despite no changes to import statements, my guess is that @GrahamGoudeau ran `go mod tidy` before committing, which should trim the dependencies imported for a module down to the required components (cleaning up any extra dependencies listed in the module files)

pkgmgmtutils/update_formulas.go
```diff
@@ -42,6 +42,10 @@ type FormulaOptions struct {
 	LinuxShaRegex   string
 	WindowsShaRegex string
 
+	// if true, open a PR even if this version is something like "x.y.z-beta1"
+	// per https://docs.brew.sh/Acceptable-Formulae#stable-versions, this is not allowed for homebrew-core
+	PublishPreRelease bool
```
kdorosh, line 47: pre-release has a different meaning for git releases. e.g., our beta releases are not marked as pre-releases. a better name for this would be `PublishUnstableRelease` or something of the like, thoughts?

pkgmgmtutils/update_formulas.go
```diff
@@ -51,11 +55,20 @@ type FormulaStatus struct {
...
+) ([]FormulaStatus, error) {
+
+	versionStruct := versionutils.GetReleaseVersionOrExitGracefully()
+	versionStr := versionStruct.String()
+	version := versionStr[1:] // skip the leading "v" in the version string
 
-	versionStr := versionutils.GetReleaseVersionOrExitGracefully().String()
-	version := versionStr[1:]
+	// we're prerelease if we have some label like "rc" or "beta"
+	isPreRelease := versionStruct.Label != ""
```
kdorosh, line 71: same thoughts about prerelease naming

go.mod
mlholland: Gotcha, thanks for the explanation.

GrahamGoudeau: In other projects, we enforce that `go mod tidy` is run prior to commit. We don't do that here just because this repo doesn't have a clear owner. In particular, it will update the go.sum file (a cryptographic hash of downloaded module content, for security) with versions of packages published for all available build platforms. That's why a lot of the changes here are additive. And in other projects we enforce that any diff to generated files (like go.sum) should fail the CI run. So I ran it out of habit here, thus this diff.

pkgmgmtutils/brew/formula_updater.go
```diff
@@ -0,0 +1,179 @@
...
+		// a version is not stable if it has a label, like "rc", "beta", etc.
+		// in either case, silently mark it as updated and continue
+		if formulaOptions.DryRun || (version.Label != "" && !formulaOptions.PublishUnstableVersion) {
+			status.Updated = true
+			formulaStatuses = append(formulaStatuses, status)
+			continue
+		}
+
+		// Suffix branch name with random number to prevent collisions in rebuilding releases
+		branchName := fmt.Sprintf("%s-%s-%d", formulaOptions.FormulaName, version, f.random.Intn(1000))
```
kdorosh, line 106: not a big fan of the collision prevention logic here, can we use the git commit hash instead?

kdorosh: or git commit sha and not push if the branch already exists?

pkgmgmtutils/brew/formula_updater_test.go
```diff
@@ -0,0 +1,338 @@
+package brew_test
```
kdorosh, line 1: we use this code to update formulas that aren't brew (e.g. fish-food), i'd recommend renaming the entire package

kdorosh: including the file names

pkgmgmtutils/utils_test.go
```diff
@@ -1,160 +0,0 @@
...
-	"io/ioutil"
-	"os"
-	"path/filepath"
-	"regexp"
-
-	. "github.com/onsi/ginkgo"
-	. "github.com/onsi/gomega"
-)
-
-var _ = Describe("package management utils", func() {
```
kdorosh, line 14: we just deleted all these tests without any replacement for them after the refactor?

GrahamGoudeau: There are maybe two that need replacing. The vast majority of this file is either covered already by my new tests or is dead code paths.

changelog/v0.16.1/stop-prerelease-brew-prs.yaml
```diff
@@ -0,0 +1,4 @@
+changelog:
+  - type: FIX
+    description: Stop opening PRs to homebrew-core for non-stable versions, which is disallowed per https://docs.brew.sh/Acceptable-Formulae#stable-versions.
+    issueLink: https://github.com/solo-io/gloo/issues/3178
```
kdorosh, line 4: `resolvesIssue: false` because the root issue still persists until this change is picked up in built bot

Issue #413, opened 2020-07-01
Allow injecting logger into stats server


marcogschmidt:
The stats server currently [builds its own logger](https://github.com/solo-io/go-utils/blob/3d5531ff239c0f28c0f685e627a175ef1c577e15/stats/stats.go#L58). This is limiting in cases where clients want to set up their own logger, but also want to allow users to change the log level at runtime using the stats server.

PR #414, opened 2020-07-01
Add a new `LogLevel` startup option to the stats server

marcogschmidt:
Add a new `LogLevel` startup option to inject the `zap.AtomicLevel` into the stats server. This way clients can build their own logger and use the stats server to change its log level at runtime, instead of having to rely on the logger built internally by the server.
BOT NOTES: 
resolves https://github.com/solo-io/go-utils/issues/413


marcogschmidt:
Can't see how this could have been caused by the change. Let's try kicking. /kick

Issue #416, opened 2020-07-14
Allow -wasm tags in releases


Sodman:
We'll have a few releases coming up which will be wasm-specific builds, built against the envoy-wasm fork of enovy proxy.

Eg we will want releases with tags like :v1.5.0-rc1-wasm, v1.5.0-wasm, v1.5.0-beta1-wasm

PR #417, opened 2020-07-14
Expand on allowed tags

Sodman:
Allowed tag examples:
- `v1.2.3-foo` (new)
- `v1.2.3-foo-bar-baz` (new)
- `v1.2.3-foo1` (no change)
- `v1.2.3-foo-bar1` (new)

(Still) not allowed:
- `1.2.3-foo1-bar`

Comparing works:
- `1.2.3-foo-bar2` > `1.2.3-foo-bar1` (because of labelVersion)
- `1.2.3-foo-bar2` < `1.2.4-foo-bar1` (because of patch version)

This is needed to release `envoy-gloo:1.15.0-wasm-rc1` and future wasm tags.

BOT NOTES: 
resolves https://github.com/solo-io/go-utils/issues/416


versionutils/version.go
```diff
@@ -15,6 +15,7 @@ import (
 const (
 	SemverNilVersionValue = "v0.0.0"
 	SemverMinimumVersion  = "v0.0.1"
+	wasmLabel             = "wasm"
```
kdorosh, line 18: can we make this more generic? i'd like to be able to support tags that have generic secondary labels, e.g.:
- 1.0.0
- 1.0.0-rc1
- 1.0.0-wasm-rc1
- 1.0.0-beta1-foobar

kdorosh: i think we should just make the label version everything that comes after the initial hyphen, as this supports arbitrary amounts of hyphens following a version and doesn't require changes to our version comparison logic

kdorosh: spoke offline, allowing labels with any amount of hyphens and stripping out any numbers to get the label version number

versionutils/version.go
```diff
@@ -271,23 +272,25 @@ func ParseVersion(tag string) (*Version, error) {
...
 func parseLabelVersion(labelAndVersion string) (string, int, error) {
-	regex := regexp.MustCompile("([a-z]+)([0-9]+)")
-	// should be like ["foo1", "foo", "1"]
-	matches := regex.FindStringSubmatch(labelAndVersion)
-	if len(matches) != 3 {
-		return "", 0, eris.Errorf("invalid label and version %s", labelAndVersion)
-	}
-	label := matches[1]
-	labelVersionToParse := matches[2]
+	numsRegex := regexp.MustCompile("[0-9]+")
```
kdorosh, line 275: should we be asserting that the number is at the end in this regex?

PR #422, opened 2020-09-18
surface resource function

mlholland:
Adds a function that surfaces the kuberesource.UnstructuredResources that back a testManifest struct. At the moment you can only indirectly inspect these through the functions available in the TestManifest interface, and I've noticed some tests in the past, as well as a test I'm currently writing, would be much simpler if I could just inspect the resources directly.
BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/3498


mlholland:
dropping - modified the gloo code to no longer need this function.

PR #423, opened 2020-10-09
Add useful PortForwardGet to cliutils

ashleywang1:
I'm using it in meshctl, and this function is super useful!

### DIFF
--- a/cliutils/port_forward.go
+++ b/cliutils/port_forward.go
@@ -1,0 +1,115 @@
+package cliutils
+
+import (
+	"context"
+	"fmt"
+	"io/ioutil"
+	"net"
+	"net/http"
+	"os"
+	"os/exec"
+	"time"
+
+	"github.com/hashicorp/go-multierror"
+	errors "github.com/rotisserie/eris"
+)
+
+// Call kubectl port-forward. Callers are expected to clean up the returned portFwd *exec.cmd after the port-forward is no longer needed.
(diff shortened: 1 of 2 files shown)


arianaw66:
Should we remove from Gloo while we're at it and have that call this instead to avoid duplicate code?

cliutils/port_forward.go
```diff
@@ -0,0 +1,115 @@
...
+func PortForwardGet(ctx context.Context, namespace string, resource string, localPort string, kubePort string, verbose bool, getPath string) (string, *exec.Cmd, error) {
+
+	/** port-forward command **/
+
+	portFwd, err := PortForward(namespace, resource, localPort, kubePort, verbose)
+	if err != nil {
+		return "", nil, err
+	}
+
+	localCtx, cancel := context.WithTimeout(ctx, time.Second*30)
```
arianaw66, line 49: This change from 3 sec to 30 sec is intentional, right?

ashleywang1:
> Should we remove from Gloo while we're at it and have that call this instead to avoid duplicate code?

I'm not sure! This would require an upgrade of go-utils, which could break something.

cliutils/port_forward.go
ashleywang1: Nope!

arianaw66: Oh sorry. I think I was looking at the PR, and someone probably changed it since.

arianaw66 (approved):

PR #435, opened 2021-02-18
Fix homebrew formula tests

npolshakova:
Empty commits were getting generated in homebrew-core by Solo-io Bot because of incorrect regex parsing and change in homebrew-core.

https://github.com/solo-io/gloo/issues/4047

### DIFF
--- a/pkgmgmtutils/formula_updater_test.go
+++ b/pkgmgmtutils/formula_updater_test.go
@@ -115,8 +115,8 @@ var _ = Describe("FormulaUpdater", func() {
 						PRDescription:   "Created by Solo-io Bot",
 						PRCommitName:    "Solo-io Bot",
 						PRCommitEmail:   "bot@solo.io",
-						VersionRegex:    `tag\s*=>\s*"v([0-9.]+)",`,
-						VersionShaRegex: `revision\s*=>\s*"(.*)"`,
+						VersionRegex:    `tag:\s*"v([0-9.]+)`,
+						VersionShaRegex: `revision:\s*"(.*)"`,
 					},
 				}
 				version := "v1.6.9"
--- a/pkgmgmtutils/formula_updater_test.go
+++ b/pkgmgmtutils/formula_updater_test.go
@@ -115,8 +115,8 @@ var _ = Describe("FormulaUpdater", func() {
 						PRDescription:   "Created by Solo-io Bot",
 						PRCommitName:    "Solo-io Bot",
 						PRCommitEmail:   "bot@solo.io",
-						VersionRegex:    `:tag\s*=>\s*"v([0-9.]+)",`,
-						VersionShaRegex: `:revision\s*=>\s*"(.*)"`,
+						VersionRegex:    `tag\s*=>\s*"v([0-9.]+)",`,
+						VersionShaRegex: `revision\s*=>\s*"(.*)"`,
 					},
 				}
 				version := "v1.6.9"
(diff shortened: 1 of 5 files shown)


pkgmgmtutils/formula_updater_test.go
```diff
@@ -115,8 +115,8 @@ var _ = Describe("FormulaUpdater", func() {
 						PRDescription:   "Created by Solo-io Bot",
 						PRCommitName:    "Solo-io Bot",
 						PRCommitEmail:   "bot@solo.io",
-						VersionRegex:    `:tag\s*=>\s*"v([0-9.]+)",`,
-						VersionShaRegex: `:revision\s*=>\s*"(.*)"`,
+						VersionRegex:    `tag\s*=>\s*"v([0-9.]+)",`,
```
sam-heilbron, line 118: is there a reason this regex is different from the regex here: https://github.com/solo-io/go-utils/pull/435/files#diff-0f08ec98106c6b85f243a4383477e6545c44dde83f05e345e80240e5484a397dR266

pkgmgmtutils/internal/formula_bytes_updater_test.go
```diff
@@ -54,4 +56,46 @@ var _ = Describe("FormulaBytesUpdater", func() {
...
+		// GitHub API docs: https://developer.github.com/v3/repos/contents/#get-contents
+		gclient := github.NewClient(nil)
+		ctx := context.Background()
+		fileContent, _, _, err := gclient.Repositories.GetContents(ctx, formula.RepoOwner, formula.RepoName, formula.Path, &github.RepositoryContentGetOptions{
+			Ref: "refs/heads/master",
+		})
+		Expect(err).To(Not(HaveOccurred()))
+
+		c, err := fileContent.GetContent()
+		Expect(err).To(Not(HaveOccurred()))
```
sam-heilbron, line 88: nit: are there any github utils that we could use to achieve this?

pkgmgmtutils/internal/formula_bytes_updater_test.go
```diff
@@ -54,4 +56,46 @@ var _ = Describe("FormulaBytesUpdater", func() {
...
+
+		b, err := internal.UpdateFormulaBytes(
+			[]byte(c),
+			"7.8.9",
+			"gitcommitsha",
+			&formula_updater_types.PerPlatformSha256{},
+			&formula_updater_types.FormulaOptions{VersionRegex: formula.VersionRegex, VersionShaRegex: formula.VersionShaRegex},
+		)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(string(b)).To(ContainSubstring(testVersion))
```
sam-heilbron, line 98: Aren't we replacing the version that matches the regex with `7.8.9`? Why aren't we comparing that it matches our replaced value?

sam-heilbron (commented):
couple of small questions, but otherwise looks good!

pkgmgmtutils/formula_updater_test.go
npolshakova: No they should be the same. The formula_updater_test doesn't called UpdateFormulaBytes so the regex here doesn't matter for the test, but it should still match.

pkgmgmtutils/internal/formula_bytes_updater_test.go
npolshakova: Yes, the GetContents gets the glooctl homebrew-core contents, and then that version is replaced with our testVersion `v7.8.9`. So the original value is not `v7.8.9`, but it gets set to that by UpdateFormulaBytes

pkgmgmtutils/internal/formula_bytes_updater_test.go
npolshakova: It looks like GetFilesFromGit can be used instead!

sam-heilbron (approved):

PR #436, opened 2021-03-02
Set level with environment variable

jahvon:
This PR updates the stats server so that the log level can be set by the environment variable that is present on start. My plan is to add this environment variable into the Helm templates so that it can be configured by the values overrides.

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/4090

### DIFF
--- a/stats/stats.go
+++ b/stats/stats.go
@@ -70,6 +70,10 @@ func StartStatsServerWithPort(startupOpts StartupOptions, addhandlers ...func(mu
 			setLevel = zapcore.WarnLevel
 		case "error":
 			setLevel = zapcore.ErrorLevel
+		case "panic":
+			setLevel = zapcore.PanicLevel
+		case "fatal":
+			setLevel = zapcore.FatalLevel
 		default:
 			setLevel = zapcore.InfoLevel
 		}
--- a/stats/stats.go
+++ b/stats/stats.go
@@ -14,6 +14,7 @@ import (
 	"go.opencensus.io/stats/view"
 	"go.opencensus.io/zpages"
 	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


stats/stats.go
```diff
@@ -60,7 +61,20 @@ func StartStatsServerWithPort(startupOpts StartupOptions, addhandlers ...func(mu
...
+		switch envLogLevel {
+		case "debug":
+			setLevel = zapcore.DebugLevel
+		case "warn":
+			setLevel = zapcore.WarnLevel
+		case "error":
+			setLevel = zapcore.ErrorLevel
+		default:
+			setLevel = zapcore.InfoLevel
+		}
```
jahvon, line 75: There are also `DPanic`, `Panic`, and `Fatal` levels but I decided to just start with these. Open to thoughts on which levels should be configurable.

stats/stats.go
```diff
@@ -143,7 +157,7 @@ func Index(w http.ResponseWriter, r *http.Request) {
 var profileDescriptions = map[string]string{
 
 	"/logging": `View \ change the log level of the program. <br/>
-
```
jahvon, line 146: Whitespace changes is just from my IDE

stats/stats.go
```diff
@@ -60,7 +61,20 @@ func StartStatsServerWithPort(startupOpts StartupOptions, addhandlers ...func(mu
...
+			setLevel = zapcore.DebugLevel
+		case "warn":
+			setLevel = zapcore.WarnLevel
+		case "error":
+			setLevel = zapcore.ErrorLevel
+		default:
+			setLevel = zapcore.InfoLevel
+		}
+		logLevel = zap.NewAtomicLevelAt(setLevel)
+	} else if startupOpts.LogLevel != nil {
```
jahvon, line 77: Another approach for this is passing in the log level from higher up in the stack as a startup option. I decided against that approach since 1) that would still be based on an env variable set through helm and 2) would result in some duplicated code across the different places where log level is configurable.

stats/stats.go
kdorosh: considering the low cost, I think we should add `Panic` and `Fatal` even though I don't think they'll really be used in practice

kdorosh (approved):

PR #440, opened 2021-03-25
Custom tag comparator

saiskee:
Adds `labelOrder` to changelogValidator. If `labelOrder` exists, then changelogValidator
      will now use the order specified in this array to tie-break tags with same versions but different labels.
      e.g. if e.g. labelOrder = [ beta, alpha, predev ], then 1.7.0-beta11 > 1.7.0-alpha5 > 1.7.0-predev9.
BOT NOTES: 
resolves https://github.com/solo-io/go-utils/issues/441

### DIFF
--- a/changelogutils/changelog_test.go
+++ b/changelogutils/changelog_test.go
@@ -28,7 +28,7 @@ var _ = Describe("ChangelogTest", func() {
 			return actualErr
 		}
 
-		It("works", func() {
+		FIt("works", func() {
 			tmpDir := mustWriteTestDir()
 			defer os.RemoveAll(tmpDir)
 			changelogDir := filepath.Join(tmpDir, changelogutils.ChangelogDirectory)
@@ -44,6 +44,8 @@ var _ = Describe("ChangelogTest", func() {
 			Expect(getProposedTag("v1.0.0-beta1", tmpDir, "v1.0.0-beta2")).To(BeNil())
 			Expect(createSubdirs(changelogDir, "v1.0.0-rc1")).To(BeNil())
 			Expect(getProposedTag("v1.0.0-beta2", tmpDir, "v1.0.0-rc1")).To(BeNil())
+			Expect(createSubdirs(changelogDir, "v1.0.0-rc2")).To(BeNil())
+			Expect(getProposedTag("v1.0.0-rc1", tmpDir, "")).NotTo(BeNil())
 
 			// add a directory without 'v' prefix, which should be parsed as invalid
 			Expect(createSubdirs(changelogDir, "1.0.0-beta3")).To(BeNil())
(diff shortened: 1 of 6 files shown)


changelogutils/changelog_test.go
```diff
@@ -44,6 +44,8 @@ var _ = Describe("ChangelogTest", func() {
 			Expect(getProposedTag("v1.0.0-beta1", tmpDir, "v1.0.0-beta2")).To(BeNil())
 			Expect(createSubdirs(changelogDir, "v1.0.0-rc1")).To(BeNil())
 			Expect(getProposedTag("v1.0.0-beta2", tmpDir, "v1.0.0-rc1")).To(BeNil())
+			Expect(createSubdirs(changelogDir, "v1.0.0-rc2")).To(BeNil())
+			Expect(getProposedTag("v1.0.0-rc1", tmpDir, "")).NotTo(BeNil())
```
saiskee, line 48: Test to show that the deprecated `changelogutils.GetProposedTag()` method (not using changelogValidator), has the issue where it throws an error if `beta1, rc1` and `rc2` changelogs exist in the same directory.

versionutils/version.go
```diff
@@ -133,6 +133,33 @@ func (v Version) IsGreaterThan(lesser Version) (bool, bool) {
...
+	for i, lbl := range labelOrder {
+		if lbl == v.Label {
+			vIndex = i
+		}
+		if lbl == lesser.Label {
+			lIndex = i
+		}
+	}
+	if vIndex == -1 || lIndex == -1 {
+		return false, false
```
kdorosh, line 156: this should be an error instead?

changelogutils/validator.go
```diff
@@ -94,6 +101,9 @@ type changelogValidator struct {
 	reader ChangelogReader
 	client githubutils.RepoClient
 	code   vfsutils.MountedRepo
+	// list of arbitrary labels whos order is used to tie-break tag comparisons between
+	// versions with different labels. Labels ordered earlier are greater.
+	labelOrder []string
```
saiskee, line 106: I'm scared to hardcode any sort of arbitrary label ordering into go-utils and leave this as an implementation detail to changelog-bot.

versionutils/version_test.go
```diff
@@ -162,6 +162,22 @@ var _ = Describe("Version", func() {
 		)
 	})
 
+	FDescribe("IsGreaterThanTagWithLabelOrder", func() {
```
kdorosh, line 165: unfocus. ci should catch this?

versionutils/version.go
saiskee: I would argue that setting `isDeterminable` to `false` works logically, rather than having to propagate an error all the way up, similar to what was being done already.

kdorosh: I can get on board with that 👍

kdorosh (approved):

PR #446, opened 2021-04-15
Test some failures

kdorosh:


kdorosh:
it still uploads with `failFast` ginkgo flag, so we're set here

PR #449, opened 2021-04-16
Changelog tooling

saiskee:
## Description
Prior to this, changelogs are rendered in Gloo Edge docs by pulling down the release notes (as markdown text) from github and allowing Hugo (our docs framework) to render them on the [changelogs docs](https://docs.solo.io/gloo-edge/latest/reference/changelog/open_source/). This PR allows for more features to these changelog docs without changing the current changelog workflow. Essentially, it parses the markdown changelog release notes into a JSON format which can be interpreted by the Hugo front-end javascript to create views such as "Chronological", "Minor release", or even a changelog diff tool, which is shown below. 

Given that all solo.io projects use the same changelog tooling currently, this tooling can be used, along with some front end code,  to generate better changelogs for all projects.

https://user-images.githubusercontent.com/1731122/118501310-91d81680-b6f6-11eb-94fa-fd24902705c5.mp4


sam-heilbron (commented):
I like this!

I could see this as its own utility package instead of being inside the changlog utils since it feels like they have unrelated responsibilities. It looks like we have a docutils that could probably be updated with this tool, since I don't believe that one is being used at all.

I think a README in this package would be useful to make it very clear to a user what this tool is, and how it's used. Also, I think it's important to note that the outputted data from this tool is the inputted data for our Hugo theme.

changelogutils/changelogdocutils/merged_release.go
```diff
@@ -0,0 +1,201 @@
...
+    // Number of versions from the most previous version to display in changelog
+    NumVersions int
+    // Maximum version changelog to display on this changelog
+    MaxVersion *Version
+    // Minimum version changelog to display on this changelog
+    MinVersion *Version
+    ProjectName,
+    RepoOwner,
+    MainRepo,
+    DependentRepo string
```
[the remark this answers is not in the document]
saiskee: The reason I chose `Main` over `Enterprise` is because this same struct is used for the `minor_release` changelog, which doesn't use the open source repo since it doesn't do any multi-repo merging.

changelogutils/changelogdocutils/minor_release.go
```diff
@@ -0,0 +1,354 @@
...
+}
+
+func (g *MinorReleaseGroupedChangelogGenerator) NewReleaseData(releases []*github.RepositoryRelease) (*ReleaseData, error) {
+	r := &ReleaseData{
+		Releases: make(map[Version]*VersionData),
+	}
+	for _, release := range releases {
+		tag, err := ParseVersion(release.GetTagName())
+		if err != nil {
+			// Release name doesn't follow proper semantic versioning, skip
```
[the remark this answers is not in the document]
saiskee: Exactly, just didn't want to break generation if that does ever happen.

changelogutils/changelogdocutils/merged_release.go
```diff
@@ -0,0 +1,201 @@
...
+    var err error
+    enterpriseReleases, err := g.GetMergedEnterpriseRelease(ctx)
+    if err != nil {
+        return "", err
+    }
+    var out struct {
+        Opts        Options
+        ReleaseData *ReleaseData
+    }
+    out.Opts = g.opts
```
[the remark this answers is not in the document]
saiskee: Options has metadata that is useful for displaying links to the github page in the front-end.

changelogutils/changelogdocutils/merged_release.go
```diff
@@ -0,0 +1,201 @@
...
+type DependencyFn func(*Version) (*Version, error)
+
+type Options struct {
+    // Number of versions from the most previous version to display in changelog
+    NumVersions int
+    // Maximum version changelog to display on this changelog
+    MaxVersion *Version
+    // Minimum version changelog to display on this changelog
+    MinVersion *Version
+    ProjectName,
```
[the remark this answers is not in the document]
saiskee: Must've left this in on accident, deleted.

changelogutils/changelogdocutils/merged_release.go
```diff
@@ -0,0 +1,201 @@
...
+
+type DependencyFn func(*Version) (*Version, error)
+
+type Options struct {
+    // Number of versions from the most previous version to display in changelog
+    NumVersions int
+    // Maximum version changelog to display on this changelog
+    MaxVersion *Version
+    // Minimum version changelog to display on this changelog
+    MinVersion *Version
```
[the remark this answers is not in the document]
saiskee: Left comments explaining the behavior of these three fields.

changelogutils/changelogdocutils/merged_release.go
```diff
@@ -0,0 +1,201 @@
...
+    }
+    out.Opts = g.opts
+    out.ReleaseData = enterpriseReleases
+    res, err := json.Marshal(out)
+    return string(res), err
+}
+
+func (g *MergedReleaseGenerator) GetMergedEnterpriseRelease(ctx context.Context) (*ReleaseData, error) {
+    ossReleases, err := NewMinorReleaseGroupedChangelogGenerator(Options{
+        RepoOwner: "solo-io",
```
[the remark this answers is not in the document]
saiskee: `solo-io` shouldn't be hardcoded here, fixed. 

As far as specifying both repos, this will cause a few extra if statements and checks that I'm okay with forgoing for the sake of simplicity. Especially since we're probably the only ones using this.

changelogutils/changelogdocutils/merged_release.go
```diff
@@ -0,0 +1,201 @@
...
+        // Build out release dependency map (enterprise -> oss) using releaseDepMap as cache
+        if g.releaseDepMap[release] == nil {
+            var (
+                dep *Version
+                err error
+            )
+            if g.DependencyFunc != nil {
+                dep, err = g.DependencyFunc(&release)
+            } else {
+                dep, err = g.GetOpenSourceDependency(&release)
```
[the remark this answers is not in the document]
saiskee: Pulled into ctor.
I'm allowing a custom impl here for two-fold reasons: 
1) testing is way easier if I can just return a dependency map in the function :P

changelogutils/changelogdocutils/merged_release.go
```diff
@@ -0,0 +1,201 @@
...
+        return nil, fmt.Errorf("unable to find go.mod file in enteprise repository")
+    }
+    content, err := files[0].GetContent()
+    if err != nil {
+        return nil, err
+    }
+    // Looks for
+    // ...repoOwner/otherRepo v1.x.x-x
+    // and captures semantic version
+    semverRegex := fmt.Sprintf("%s/%s\\s+(v((([0-9]+)\\.([0-9]+)\\.([0-9]+)(?:-([0-9a-zA-Z-]+(?:\\.[0-9a-zA-Z-]+)*))?)(?:\\+([0-9a-zA-Z-]+(?:\\.[0-9a-zA-Z-]+)*))?))",
```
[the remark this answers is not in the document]
saiskee: pulled out into it's own func.

changelogutils/changelogdocutils/minor_release.go
```diff
@@ -0,0 +1,354 @@
...
+	generator *MinorReleaseGroupedChangelogGenerator
+}
+
+// Contains Changelog enterpriseNotes for Individual openSourceReleases
+// ChangelogNotes is a map of individiual openSourceReleases to enterpriseNotes
+// e.g. v1.2.5-beta3 -> ChangelogNotes, v1.4.0 -> ChangelogNotes
+type VersionData struct {
+	ChangelogNotes map[Version]*ChangelogNotes
+	generator      *MinorReleaseGroupedChangelogGenerator
+}
```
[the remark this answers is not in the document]
saiskee: Deleted generator, wasn't being used!

changeloggenutils/minor_release.go
```diff
@@ -0,0 +1,353 @@
...
+ - v1.8.2
+ - v1.8.1
+ v1.7
+ - v1.7.1
+ - v1.7.0-beta9
+ ...
+ */
+func NewMinorReleaseGroupedChangelogGenerator(opts Options, client *github.Client) *MinorReleaseGroupedChangelogGenerator {
+	if opts.NumVersions == 0{
+		opts.NumVersions = math.MaxInt64
```
sam-heilbron, line 35: Can we change the default to 1? I'm just afraid of someone forgetting to set the NumVersions option, and then we pull down tons of release notes

changeloggenutils/minor_release.go
```diff
@@ -0,0 +1,353 @@
...
+	return r, nil
+}
+
+func (g *MinorReleaseGroupedChangelogGenerator) isBetweenMinAndMaxVersions(version Version) bool{
+	res := true
+	if g.opts.MinVersion != nil{
+		res = version.MustIsGreaterThanOrEqualTo(*g.opts.MinVersion)
+	}
+	if g.opts.MaxVersion != nil {
+		res = res && !version.MustIsGreaterThanOrEqualTo(*g.opts.MaxVersion)
```
sam-heilbron, line 131: should this be is less than or equal to?

githubutils/repo.go
```diff
@@ -33,6 +34,8 @@ const (
 
 	CONTENT_TYPE_FILE      = "file"
 	CONTENT_TYPE_DIRECTORY = "dir"
+
+	MAX_GITHUB_RELEASES_PER_PAGE = 100
```
sam-heilbron, line 38: Why do we change the max release per page from 10,000,000 (https://github.com/solo-io/go-utils/pull/449/files#diff-18395ae6cd3dd026be8823a5785e272b7b12f3214e02169d1039ff2f32a48767L164) to 100?

changeloggenutils/merged_release.go
```diff
@@ -0,0 +1,252 @@
...
+    if err != nil {
+        return nil, err
+    }
+
+    return getPkgVersionFromGoMod(content, g.opts.RepoOwner, g.opts.DependentRepo)
+}
+
+// HOF that returns a Dependency function. The returned function generates
+// changelogs faster than the default but requires passing in the github token.
+func GetOSDependencyFunc(repoOwner, enterpriseRepo, osRepo, githubToken string) (DependencyFn, error){
```
sam-heilbron, line 219: How much faster is this vs the alternate depedencyFn?

changeloggenutils/merged_release.go
```diff
@@ -0,0 +1,252 @@
...
+    osSorted := osReleases.GetReleasesSorted()
+    for _, release := range enterpriseSorted {
+        // Build out release dependency map (enterprise -> oss) using releaseDepMap as cache
+        if g.releaseDepMap[release] == nil {
+            var (
+                dep *Version
+                err error
+            )
+            dep, err = g.DependencyFunc(&release)
+            if err != nil {
```
sam-heilbron, line 114: Should this error instead of continuing? This means that we were unable to find the dependency for a particular release. I assume this would only happen if (A) we had a bug in the function looking for the dep or (B) provided incorrect options for the main and dependent repos. Either way, it feels like we should fail loudly since something has been misconfigured by the user. What do you think?

sam-heilbron (commented):
How do we generate our open source docs? Is there a way to rely on this same tool to generate them? I'm picturing the option to provide a repo (for open source) or a repo and a dependent repo (for enterprise). That way all our docs are generated from the same tool.

changeloggenutils/minor_release.go
saiskee: By default, we want to pull down all release notes, as is currently being done in our changelogs. If we set default to 1, that will only pull down one release. `NumVersions` is maybe for later stage products with thousands of releases.

githubutils/repo.go
saiskee: https://docs.github.com/en/rest/overview/resources-in-the-rest-api#pagination
> For some resources, you can also set a custom page size up to 100 with the per_page parameter

changeloggenutils/merged_release.go
saiskee: Since changelogs are generated at release time, I want to decrease the chance of a release failing due to user error. The worst case here is that open source dependencies don't show up for an enterprise releases' changelogs. I don't feel too strongly about it, but if a version tag is messed up for some reason, I'd rather have some other validation pick up that error than fail changelog generation completely.

changeloggenutils/merged_release.go
saiskee: It's a matter of how many versions have been released. This clones an entire repo in memory, which takes about 10 seconds, then the look up of each go.mod file for ~200 versions takes less than 1 second. 
The default method queries github API on every request, so for ~100 versions it takes 10 seconds, but the query time per version is much higher than the in-memory func.

changeloggenutils/minor_release.go
```diff
@@ -0,0 +1,353 @@
...
+	if opts.NumVersions == 0{
+		opts.NumVersions = math.MaxInt64
+	}
+	return &MinorReleaseGroupedChangelogGenerator{
+		opts:      opts,
+		Client:    client,
+	}
+}
+
+func (g *MinorReleaseGroupedChangelogGenerator) GenerateJSON(ctx context.Context) (string, error) {
```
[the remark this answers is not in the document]
saiskee: This the entry point that will be called by the repo's.

changeloggenutils/merged_release.go
```diff
@@ -0,0 +1,252 @@
...
+    }
+    enterpriseReleases, err := NewMinorReleaseGroupedChangelogGenerator(g.opts, g.client).
+        GetReleaseData(ctx)
+    if err != nil {
+        return nil, err
+    }
+    return g.MergeEnterpriseReleaseWithOS(ctx, enterpriseReleases, ossReleases)
+}
+
+func (g *MergedReleaseGenerator) MergeEnterpriseReleaseWithOS(ctx context.Context, enterpriseReleases, osReleases *ReleaseData) (*ReleaseData, error) {
```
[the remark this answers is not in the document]
saiskee: Nope! good catch.

versionutils/version.go
```diff
@@ -337,3 +342,21 @@ func MatchesRegex(tag string) bool {
 func GetImageVersion(version *Version) string {
 	return version.String()[1:]
 }
+
+func Index(versions []Version, v Version) int {
```
sam-heilbron, line 346: I think this is unused

stringutils/strings.go
```diff
@@ -46,6 +46,15 @@ func KeysAndValues(m map[string]string) ([]string, []string) {
 	return keys, values
 }
 
+func Keys(m map[string]interface{}) []string {
```
sam-heilbron, line 49: nit: can we reference this function in the KeysAndValues function above?

versionutils/version.go
saiskee: If we can leave this in, it may be useful in the future?

sam-heilbron (approved):

PR #451, opened 2021-05-06
Build logger when an env var is set

jahvon:
When setting the stat server log level via an environment variable. previously the log config was not being built so logs would not be produced. This also prevented logs from being produced when set via the admin api after the server started without a log config. This fix builds the log config so that the log level is applied and logs are produced.

Tested this locally with a gloo image that pulled in this change and a defined the env var.

### DIFF
--- a/stats/stats.go
+++ b/stats/stats.go
@@ -78,6 +78,13 @@ func StartStatsServerWithPort(startupOpts StartupOptions, addhandlers ...func(mu
 			setLevel = zapcore.InfoLevel
 		}
 		logLevel = zap.NewAtomicLevelAt(setLevel)
+
+		logConfig := zap.NewProductionConfig()
+		logConfig.Level = logLevel
+		logger, logErr := logConfig.Build()
+		if logErr == nil {
+			contextutils.SetFallbackLogger(logger.Sugar())
+		}
 	} else if startupOpts.LogLevel != nil {
 		logLevel = *startupOpts.LogLevel
 	} else {
--- a/changelog/v0.21.5/log-level-fix.yaml
+++ b/changelog/v0.21.5/log-level-fix.yaml
@@ -1,0 +1,5 @@
+changelog:
+- type: FIX
+  description: Fix setting of stat server log level when using the LOG_LEVEL environment variable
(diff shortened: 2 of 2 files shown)


saiskee (approved):
Looks good to me! Going to relieve a lot of customer pain for sure :)

stats/stats.go
```diff
@@ -78,6 +78,13 @@ func StartStatsServerWithPort(startupOpts StartupOptions, addhandlers ...func(mu
...
 		logLevel = zap.NewAtomicLevelAt(setLevel)
+
+		logConfig := zap.NewProductionConfig()
+		logConfig.Level = logLevel
+		logger, logErr := logConfig.Build()
+		if logErr == nil {
+			contextutils.SetFallbackLogger(logger.Sugar())
+		}
 	} else if startupOpts.LogLevel != nil {
 		logLevel = *startupOpts.LogLevel
```
sam-heilbron, line 89: Is there a reason this case doesn't need the same `zap.NewProductionConfig` lines that were added?

stats/stats.go
```diff
@@ -78,6 +78,13 @@ func StartStatsServerWithPort(startupOpts StartupOptions, addhandlers ...func(mu
 			setLevel = zapcore.InfoLevel
 		}
 		logLevel = zap.NewAtomicLevelAt(setLevel)
+
+		logConfig := zap.NewProductionConfig()
```
sam-heilbron, line 82: Has this issue always existed? I thought the ticket mentioned this was just for 1.7.x?

jahvon: I added the setting of log level via env variable in 1.7 but think I just missed the step of building that when it was initially added. That's likely why it's only seen for 1.7 (when users are trying to set the level via env var)

stats/stats.go
jahvon: I checked Gloo for this as well. Looks like the accesslogger component is the only thing that is setting a loglevel through the startup options. In that case, it is also creating it's own logger in `contextutils` so I'm guessing if we define it this way then it is expected that the logger is set outside of this scope.

PR #453, opened 2021-06-21
Add option for using cached github repository releases

saiskee:
Adds option for changelog generator to use cached github repository releases rather than fetching them.


changeloggenutils/minor_release.go
```diff
@@ -21,46 +22,58 @@ type MinorReleaseGroupedChangelogGenerator struct {
...
 	out.ReleaseData = releaseData
 	res, err := json.Marshal(out)
 	return string(res), err
 }
 
-func (g *MinorReleaseGroupedChangelogGenerator) GetReleaseData(ctx context.Context) (*ReleaseData, error) {
-	releases, err := githubutils.GetAllRepoReleasesWithMax(ctx, g.Client, g.opts.RepoOwner, g.opts.MainRepo, g.opts.NumVersions)
-	if err != nil {
-		return nil, err
+func (g *MinorReleaseGroupedChangelogGenerator) GetReleaseData(ctx context.Context, cachedReleases []*github.RepositoryRelease) (*ReleaseData, error) {
```
saiskee, line 65: Updated this function to return `cachedReleases` if it exists, else to fetch the releases itself.

Makefile
```diff
@@ -5,4 +5,13 @@
...
+	git config core.hooksPath .githooks
+
+.PHONY: install-go-tools
+install-go-tools:
+	go install golang.org/x/tools/cmd/goimports
+
+.PHONY: go-fmt
+go-fmt: install-go-tools
+	gofmt -w .
+	goimports -w .
```
jenshu, line 17: doesn't goimports [also run gofmt](https://pkg.go.dev/golang.org/x/tools/cmd/goimports?utm_source=godoc)?
i think we can remove the gofmt line

.github/workflows/codeformat.yaml
```diff
@@ -0,0 +1,46 @@
+name: pull_request
```
saiskee, line 1: Adds codeformat (`gofmt and goimports`) check to CI

ci/check-code-format.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+set -ex
+
+set +e
```
ben-taussig-solo, line 5: Is there a reason to call both `set +e` and `set -ex`? Seems like this might be equivalent to `set -x` (assuming set +e is default behavior)

ben-taussig-solo (approved):
Looks good to me

PR #455, opened 2021-07-19
Add security scan utilities

saiskee:
BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/5048


securityscanutils/securityscan.go
```diff
@@ -0,0 +1,284 @@
...
+	if err != nil {
+		return eris.Wrap(err, "error initializing github client")
+	}
+	for _, repo := range s.Repos {
+		opts := repo.Opts
+		allReleases, err := githubutils.GetAllRepoReleases(ctx, client, repo.Owner, repo.Repo)
+		if err != nil {
+			return eris.Wrapf(err, "unable to fetch all github releases for github.com/%s/%s", repo.Owner, repo.Repo)
+		}
+		githubutils.SortReleasesBySemver(allReleases)
```
sam-heilbron, line 91: we call sort and then filteredReleases. (1) are we assured that the ordering is maintained in the new list? (2) we could reduce the number of items that are sorted by sorting only on the filtered releases.

securityscanutils/securityscan.go
```diff
@@ -0,0 +1,284 @@
...
+		// Filter releases by version constraint provided
+		filteredReleases := githubutils.FilterReleases(allReleases, opts.VersionConstraint)
+		markdownTplFile, err := GetTemplateFile(MarkdownTrivyTemplate)
```
sam-heilbron, line 94: we create tmp files from this fuction. Per https://github.com/golang/go/blob/master/src/io/ioutil/tempfile.go#L20 `It is the caller's responsibility to remove the file when no longer needed.`. Can we ensure that we cleanup this tmp file after using it?

securityscanutils/securityscan.go
```diff
@@ -0,0 +1,284 @@
...
+		_, err = RunTrivyScan(imageWithRepo, version, markdownTplFile, output)
+		if err != nil {
+			return eris.Wrapf(err, "error running image scan on image %s", imageWithRepo)
+		}
+	}
+	return nil
+}
+
+func (r *SecurityScanRepo) RunGithubSarifScan(versionToScan *semver.Version, sarifTplFile string) error {
+	images, err := r.GetImagesToScan(versionToScan)
```
sam-heilbron, line 144: Can we short-circuit this logic even more to check if r.Options.UploadCodeScanToGithub is true? That way we don't even perform all the extra scans if we don't need them

securityscanutils/securityscan.go
```diff
@@ -0,0 +1,284 @@
...
+func (r *SecurityScanRepo) UploadSecurityScanToGithub(fileName, versionTag string) error {
+	githubRepoApiUrl := fmt.Sprintf("https://api.github.com/repos/%s/%s", r.Owner, r.Repo)
+	githubToken, err := osutils.GetEnvE("GITHUB_TOKEN")
+	if err != nil {
+		return err
+	}
+	resp, err := req.Get(githubRepoApiUrl+"/git/refs/tags/v"+versionTag, req.Header{"Authorization": "token " + githubToken})
+	if err != nil {
+		return eris.Wrapf(err, "Unable to get commit for version v%s", versionTag)
+	}
```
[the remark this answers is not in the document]
saiskee: I'm keeping this in securityscanutils for two reasons:
1. githubutils uses the go-github api client, and after thoroughly reading through the github go-github library, I'm unable to find the equivalent client code for this endpoint, even in newer versions of go-github client.
2. This function is only ever going to be used in the context of security scans.

securityscanutils/securityscan.go
```diff
@@ -0,0 +1,284 @@
...
+	header := req.Header{
+		"Authorization": fmt.Sprintf("token %s", githubToken),
+		"Content-Type":  "application/json",
+	}
+	res, err := req.Post(githubRepoApiUrl+"/code-scanning/sarifs", req.BodyJSON(sarifMetadata), header)
+	fmt.Println(res.String())
+	if err != nil || res.Response().StatusCode != 200 {
+		return eris.Wrapf(err, "error uploading sarif file to github, response: \n%s", res)
+	}
+	return nil
```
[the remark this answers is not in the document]
saiskee: I'm keeping this in securityscanutils for two reasons:

1. githubutils uses the go-github api client, and after thoroughly reading through the github go-github library, I'm unable to find the equivalent client code for this endpoint, even in newer versions of go-github client.
2. This function is only ever going to be used in the context of security scans.

securityscanutils/securityscan.go
```diff
@@ -0,0 +1,284 @@
...
+		return eris.Wrapf(err, "error reading sarif file %s", fileName)
+	}
+	var sarifFile bytes.Buffer
+	w := gzip.NewWriter(&sarifFile)
+	_, err = w.Write(b)
+	if err != nil {
+		return eris.Wrap(err, "error writing gzip file")
+	}
+	w.Close()
+	if len(shaResp.Sha) != 40 {
```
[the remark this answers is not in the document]
saiskee: A couple times when I was getting rate limited, the api sent a 200 OK response but left the SHA empty. I'm not sure why, and I haven't hit this recently, but I want to keep this check here to prevent a really annoying debugging session in the future if the error pops up more.

securityscanutils/securityscan.go
saiskee: Thanks for pointing this out! I was also calling this in a for loop, which would create an unnecessary amount of temp files.

securityscanutils/securityscan.go
```diff
@@ -0,0 +1,284 @@
...
+	}
+	version := versionToScan.String()
+	outputDir := path.Join(r.Opts.OutputDir, r.Repo, "sarif_results", version)
+	err = osutils.CreateDirIfNotExists(outputDir)
+	if err != nil {
+		return err
+	}
+	for _, image := range images {
+		imageWithRepo := fmt.Sprintf("%s/%s:%s", r.Opts.ImageRepo, image, version)
+		fileName := fmt.Sprintf("%s_cve_report.sarif", image)
```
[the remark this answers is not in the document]
saiskee: i don't see any harm in keeping it around, especially if someone were to generate this locally. I've already shown that this will be part of the output directory [here](https://github.com/solo-io/go-utils/blob/108a37f922d9598980952134bf659310d78ff694/securityscanutils/securityscan.go#L44). I estimate all the sarif files will take up ~2 MB of memory after even a 100 scans so it shouldn't be bad as far as resource usage.

osutils/utils.go
```diff
@@ -0,0 +1,39 @@
...
+		return "", MissingEnvVarError(envVar)
+	}
+	return env, nil
+}
+
+/**
+  Creates dir if it doesn't already exist
+  If the dir does exist, does nothing.
+*/
+func CreateDirIfNotExists(path string) error {
```
sam-heilbron, line 31: from MkdirAll comment: `// If path is already a directory, MkdirAll does nothing and returns nil`...do we need this function or can we rely on MkdirAll?

osutils/utils.go
```diff
@@ -0,0 +1,39 @@
...
+var (
+	MissingEnvVarError = func(envVar string) error {
+		return fmt.Errorf("Missing %s environment variable", envVar)
+	}
+)
+
+/**
+  Returns error if the environment variable is empty
+*/
+func GetEnvE(envVar string) (string, error) {
```
sam-heilbron, line 19: Nit: we generally prefix with `must` if an error is returned in case of a missing value

securityscanutils/securityscan.go
```diff
@@ -0,0 +1,281 @@
...
+	// e.g. https://github.com/solo-io/gloo/security/code-scanning
+	UploadCodeScanToGithub bool
+}
+
+// Main method to call on SecurityScanner which generates .md and .sarif files
+// in OutputDir as defined above per repo. If UploadCodeScanToGithub is true,
+// sarif files will be uploaded to the repository's code-scanning endpoint.
+func (s *SecurityScanner) GenerateSecurityScans(ctx context.Context) error {
+
+	client, err := githubutils.GetClient(ctx)
```
sam-heilbron, line 81: Note: this will fail if GITHUB_TOKEN env variable is not set. Could be good to call out in security scanning docs that this is required

securityscanutils/securityscan.go
```diff
@@ -0,0 +1,281 @@
...
+				}
+			}
+		}
+
+	}
+	return nil
+}
+
+func (r *SecurityScanRepo) RunMarkdownScan(versionToScan *semver.Version, markdownTplFile string) error {
+	images, err := r.GetImagesToScan(versionToScan)
```
sam-heilbron, line 129: Do we need to filter these images? I thought executing `filteredReleases := githubutils.FilterReleases(allReleases, opts.VersionConstraint)` which we do earlier would protect us against scanning images that don't fit the constraint

securityscanutils/securityscan.go
```diff
@@ -0,0 +1,281 @@
...
+			err = r.UploadSecurityScanToGithub(output, version)
+			if err != nil {
+				return eris.Wrapf(err, "error uploading security scan results sarif to github for version %s", version)
+			}
+		}
+	}
+	return nil
+}
+
+func (r *SecurityScanRepo) GetImagesToScan(versionToScan *semver.Version) ([]string, error) {
```
sam-heilbron, line 180: commented above, but wondering what the difference between this function is, and the github_utils.FilterReleases is

securityscanutils/securityscan.go
```diff
@@ -0,0 +1,281 @@
...
+		CommitSha: sha,
+		Sarif:     base64.StdEncoding.EncodeToString(sarifFile.Bytes()),
+	}
+	header := req.Header{
+		"Authorization": fmt.Sprintf("token %s", githubToken),
+		"Content-Type":  "application/json",
+	}
+	url := fmt.Sprintf("https://api.github.com/repos/%s/%s/code-scanning/sarifs", r.Owner, r.Repo)
+	res, err := req.Post(url, req.BodyJSON(sarifMetadata), header)
+	fmt.Println(res.String())
```
sam-heilbron, line 269: remove?

securityscanutils/securityscan.go
```diff
@@ -0,0 +1,281 @@
...
+		Ref:       fmt.Sprintf("refs/tags/v%s", versionTag),
+		CommitSha: sha,
+		Sarif:     base64.StdEncoding.EncodeToString(sarifFile.Bytes()),
+	}
+	header := req.Header{
+		"Authorization": fmt.Sprintf("token %s", githubToken),
+		"Content-Type":  "application/json",
+	}
+	url := fmt.Sprintf("https://api.github.com/repos/%s/%s/code-scanning/sarifs", r.Owner, r.Repo)
+	res, err := req.Post(url, req.BodyJSON(sarifMetadata), header)
```
sam-heilbron, line 268: do we need call `defer req.Body.Close()`?

securityscanutils/securityscan.go
```diff
@@ -0,0 +1,281 @@
...
+		return eris.Wrapf(err, "error reading sarif file %s", fileName)
+	}
+	var sarifFile bytes.Buffer
+	gzipWriter := gzip.NewWriter(&sarifFile)
+	_, err = gzipWriter.Write(sarifFileBytes)
+	if err != nil {
+		return eris.Wrap(err, "error writing gzip file")
+	}
+	gzipWriter.Close()
+	if len(sha) != 40 {
```
sam-heilbron, line 255: what would cause this case to occur?

securityscanutils/trivy_templates.go
```diff
@@ -0,0 +1,137 @@
...
+func GetTemplateFile(trivyTemplate string) (string, error) {
+	f, err := ioutil.TempFile("", "")
+	if err != nil {
+		return "", eris.Wrap(err, "Unable to create temporary file to write template to")
+	}
+	_, err = f.Write([]byte(trivyTemplate))
+	if err != nil {
+		return "", eris.Wrapf(err, "Unable to write template to file %s", f.Name())
+	}
+	return f.Name(), nil
```
sam-heilbron, line 136: are we intentionally leaving the file open for writing?

go.mod
```diff
@@ -20,6 +21,7 @@ require (
 	github.com/grpc-ecosystem/go-grpc-middleware v1.0.0
 	github.com/hashicorp/go-multierror v1.0.0
 	github.com/hinshun/vt10x v0.0.0-20180809195222-d55458df857c
+	github.com/imroc/req v0.3.0
```
sam-heilbron, line 24: what benefits does this library provide outside the http library?

osutils/utils.go
saiskee: https://www.journal.petrausch.info/post/2020/05/must-pattern/
> In conclusion, use the prefix Must if your method or function panics. Be careful if an API provides a Must function. Expect that function to panic. 

I don't want this function to panic, but return an error.

securityscanutils/securityscan.go
saiskee: FilterReleases uses the VersionConstraint defined [here](https://github.com/solo-io/go-utils/blob/108a37f922d9598980952134bf659310d78ff694/securityscanutils/securityscan.go#L66). This `GetImagesToScan` uses the [ImagesPerVersion](https://github.com/solo-io/go-utils/blob/108a37f922d9598980952134bf659310d78ff694/securityscanutils/securityscan.go#L62) map to figure out which images belong to which version. We are not filtering anything here, but rather retrieving which images actually belong to this verison.

securityscanutils/securityscan.go
saiskee: Going to keep this as a log.

securityscanutils/securityscan.go
saiskee: The imroc library does this for us.

securityscanutils/securityscan.go
saiskee: https://github.com/solo-io/go-utils/pull/455#discussion_r673579465

securityscanutils/trivy_templates.go
saiskee: All writing is done here, I should close this file in this method.

go.mod
saiskee: Http requests in less lines of code. Golang is known for extremely verbose http request code (for example, writing to a request body for a POST request requires opening up a Writer, then writing a string to it). The imroc library does this in less lines: 
```go
	res, err := req.Post(url, req.BodyJSON(sarifMetadata), header)
```

We could implement our own httputils. But I settled for the quicker approach of using a library.

securityscanutils/securityscan.go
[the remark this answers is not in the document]
sam-heilbron: ahh makes sense. perhaps we could ensure that a 200 is returned by the request, but good for now!

sam-heilbron (approved):

PR #456, opened 2021-07-23
Create vulnerability Github Issues on Trivy scan

saiskee:
Adds option to Security Scanner to create github issues for each image that a vulnerability is found in.
      If an issue already exists for that image, it will update the issue with the most recent vulnerability scan results. 

This is especially useful for reporting vulnerabilities in private repositories, because the "Code scanning results" tab for private repositories is a paid-feature.

Example generated issue: https://github.com/solo-io/solo-projects/issues/2467

### DIFF
--- a/securityscanutils/README.md
+++ b/securityscanutils/README.md
@@ -52,6 +52,8 @@ func main() {
                     // endpoint, e.g. https://github.com/solo-io/gloo/security/code-scanning
                     // read more here: https://docs.github.com/en/rest/reference/code-scanning
 					UploadCodeScanToGithub: true,
+					// Opens/Updates Github Issue for each image a vulnerability is found in
+                    CreateGithubIssuePerImageVulnerability: true,
 				},
 			},
 		},
--- a/securityscanutils/securityscan.go
+++ b/securityscanutils/securityscan.go
@@ -33,6 +33,8 @@ type SecurityScanRepo struct {
 	Repo  string
 	Owner string
 	Opts  *SecurityScanOpts
+
+	allGithubIssues []*github.Issue
 }
 
 type SecurityScanOpts struct {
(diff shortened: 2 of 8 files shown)


securityscanutils/README.md
```diff
@@ -52,6 +52,8 @@ func main() {
                     // endpoint, e.g. https://github.com/solo-io/gloo/security/code-scanning
                     // read more here: https://docs.github.com/en/rest/reference/code-scanning
 					UploadCodeScanToGithub: true,
+					// Opens/Updates Github Issue for each image a vulnerability is found in
```
MitchAman, line 55: Formatting looks off here?

securityscanutils/securityscan.go
```diff
@@ -96,18 +109,24 @@ func (s *SecurityScanner) GenerateSecurityScans(ctx context.Context) error {
...
-		allReleases, err := githubutils.GetAllRepoReleases(ctx, client, repo.Owner, repo.Repo)
+		allReleases, err := githubutils.GetAllRepoReleases(ctx, s.githubClient, repo.Owner, repo.Repo)
 		if err != nil {
 			return eris.Wrapf(err, "unable to fetch all github releases for github.com/%s/%s", repo.Owner, repo.Repo)
 		}
 		// Filter releases by version constraint provided
 		filteredReleases := githubutils.FilterReleases(allReleases, opts.VersionConstraint)
 		githubutils.SortReleasesBySemver(filteredReleases)
+		if repo.Opts.CreateGithubIssuePerImageVulnerability {
+			repo.allGithubIssues, err = githubutils.GetAllIssues(ctx, s.githubClient, repo.Owner, repo.Repo, &github.IssueListByRepoOptions{})
```
MitchAman, line 120: (1) Can we use these options to specify a label or something that would reduce the number of results returned?  IIUC, we're fetching all the issues just so we can check if an issue has already been created for this vulnerability? If we're creating a new vulnerability issue we could apply the label during creation, then use it to reduce the size of this query.

(2) Does `githubutils.GetAllIssues` return only open issues, or is that another option we could potentially specify here?

securityscanutils/securityscan.go
```diff
@@ -266,10 +306,47 @@ func (r *SecurityScanRepo) UploadSecurityScanToGithub(fileName, versionTag strin
...
+	issueTitle := fmt.Sprintf("Security Alert: %s", image)
+	trivyLabels := &[]string{"trivy", "vulnerability"}
+	markdownScan, err := ioutil.ReadFile(markdownScanFilePath)
+	if err != nil {
+		return eris.Wrapf(err, "error reading file %s", markdownScanFilePath)
+	}
+	issueRequest := &github.IssueRequest{
+		Title:  github.String(issueTitle),
+		Body:   github.String(string(markdownScan)),
+		Labels: trivyLabels,
```
MitchAman, line 329: Ah, so it looks like we do already apply a label during creation

securityscanutils/securityscan.go
```diff
@@ -266,10 +306,47 @@ func (r *SecurityScanRepo) UploadSecurityScanToGithub(fileName, versionTag strin
...
+	}
+	createNewIssue := true
+
+	for _, issue := range r.allGithubIssues {
+		// If issue already exists, update existing issue with new security scan
+		if strings.Contains(issue.GetTitle(), issueTitle) {
+			// Only create new issue if issue does not already exist
```
MitchAman, line 336: Only create new issue if (issue does not already exist **|| existing issue is closed)**

Not sure if it's possible we'd have duplicate titles, but if we somehow do, we don't want to update a closed issue and not catch wind of the vulnerability

MitchAman (commented):
Looks good!  Feel free to push back on any of my suggestions, mostly just tossing ideas out there

securityscanutils/README.md
saiskee: This is how the fomratting looks in my IDE:
image even after pulling from this branch. Might be a GH rendering issue?

securityscanutils/securityscan.go
saiskee: 1) Good idea! 👍 

2) By default it returns only open issues :)

securityscanutils/securityscan.go
saiskee: We only get open issues from github, so if the existing issues don't contain the title, it doesn't exist or is closed

MitchAman (approved):

PR #457, opened 2021-07-26
Securityscanutils - Creates only *one* github issue per version scanned

saiskee:


securityscanutils/securityscan.go
```diff
@@ -169,13 +170,23 @@ func (r *SecurityScanRepo) RunMarkdownScan(ctx context.Context, client *github.C
...
+		}
+
+		if vulnFound {
+			vulnerabilityMd += fmt.Sprintf("# %s\n\n %s\n\n", imageWithRepo, trivyScanMd)
+		}
+
+	}
+	// Create / Update Github issue for the repo if a vulnerability is found
+	// and CreateGithubIssuePerImageVulnerability is set to true
+	if r.Opts.CreateGithubIssuePerImageVulnerability {
```
jenshu, line 185: should this be renamed to `CreateGithubIssuePerImage`? or do we still allow the option of creating a separate issue per vulnerability?

saiskee: Ah good catch, will rename.

jenshu (approved):

PR #461, opened 2021-08-02
ToHashSet and MapSlice stringutils

saiskee:


stringutils/strings.go
```diff
@@ -66,3 +66,26 @@ func Unique(stringSlice []string) []string {
...
+		set[s] = true
+	}
+	return set
+}
+
+// returns a new array where f is applied on all values in stringSlice
+// usage:
+// fmt.Println(MapSlice([]string{"abc","def"}, strings.ToUpper)
+// result: []string{"ABC", "DEF}
+func MapSlice(stringSlice []string, f func(string) string) []string {
```
kdorosh, line 85: perhaps `MapStringSlice` in case we want to add future map functions for other types?

stringutils/strings.go
```diff
@@ -66,3 +66,26 @@ func Unique(stringSlice []string) []string {
 	}
 	return list
 }
+
+func ToHashSet(stringSlice []string) map[string]bool {
```
kdorosh, line 70: same nit for name here

kdorosh (approved):

PR #462, opened 2021-08-09
Support Trivy Scan Retries

sam-heilbron:
- Introduce a RepositoryReleasesPredicate, allowing clients to define the filter behavior for github repository releases. Ideally, we could configure this in the API request, but unfortunately that is not supported at the moment.
- Add retries to our Trivy scanning logic. We noticed a flake in a previous run: https://github.com/solo-io/gloo/runs/3280618311?check_suite_focus=true, that we hope will be resolved by adding some retries, though we are not sure.

### DIFF
--- a/securityscanutils/securityscan.go
+++ b/securityscanutils/securityscan.go
@@ -7,21 +7,22 @@ import (
 	"encoding/base64"
 	"fmt"
 	"io/ioutil"
+	"math"
 	"os"
 	"os/exec"
 	"path"
 	"strconv"
 	"strings"
 
+	"github.com/Masterminds/semver/v3"
+
 	"github.com/google/go-github/v32/github"
 	"github.com/solo-io/go-utils/osutils/executils"
 
-	"github.com/Masterminds/semver/v3"
 	"github.com/imroc/req"
 	"github.com/rotisserie/eris"
 	"github.com/solo-io/go-utils/githubutils"
 	"github.com/solo-io/go-utils/log"
-	"github.com/solo-io/go-utils/osutils"
 )
 
 type SecurityScanner struct {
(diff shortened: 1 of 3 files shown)


securityscanutils/securityscan.go
```diff
@@ -244,6 +250,43 @@ func (r *SecurityScanRepo) GetImagesToScan(versionToScan *semver.Version) ([]str
...
+// github.RepositoryRelease artifacts should be included in the bulk security scan
+// At the moment, the two requirements are that:
+// 1. The release is not a pre-release or draft
+// 2. The release matches the configured version constraint
+type SecurityScanRepositoryReleasePredicate struct {
+	versionConstraint *semver.Constraints
+}
+
+func (s *SecurityScanRepositoryReleasePredicate) Apply(release *github.RepositoryRelease) bool {
+	if release.GetPrerelease() || release.GetDraft() {
```
sam-heilbron, line 269: This is net new. In previous runs of the security scan, https://github.com/solo-io/gloo/runs/3283222189?check_suite_focus=true for example, there are logs : 
```
"Mon, 09 Aug 2021 18:06:21 UTC: /home/runner/go/pkg/mod/github.com/solo-io/go-utils@v0.21.14/securityscanutils/securityscan.go:276"	WARNING:
image "quay.io/solo-io/access-logger:1.6.20" not found for version "1.6.20"
```

This happens when we attempt to execute a scan against an image that wasn't found. To avoid this warning, we filter releases to exclude draft and pre-releases.

sam-heilbron: I added a comment https://github.com/solo-io/go-utils/pull/462/files#diff-332c2f04abe6a9bb1c44001932ede0a429978bbb8c4efe0b78de042e6eee0654R322 around where we log that error, identifying that we should no longer get the warning (since we're filtering out those images) but that we'll keep it in to be warned of other situations where we can't find the image

saiskee (approved):

PR #465, opened 2021-09-14
Deprecate backoff functions with typo

jehawley:
Reviewers should confirm that this is not a breaking change before approving.

### DIFF
--- a/contextutils/backoff.go
+++ b/contextutils/backoff.go
@@ -18,30 +18,38 @@ const (
 	jitter = 0.2
 )
 
-type ExponentioalBackoff struct {
+type ExponentialBackoff struct {
 	MaxRetries  uint
 	MaxDuration *time.Duration
 	MaxDelay    *time.Duration
 }
 
+// Deprecated - use ExponentialBackoff instead
+type ExponentioalBackoff = ExponentialBackoff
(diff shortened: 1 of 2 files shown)


contextutils/backoff.go
```diff
@@ -18,38 +18,46 @@ const (
...
 }
 
+// Deprecated - use ExponentialBackoff instead
+type ExponentioalBackoff = ExponentialBackoff
+
 type Backoff interface {
 	Backoff(ctx context.Context, f func(ctx context.Context) error) error
 }
 
+// Deprecated - use NewExponentialBackoff instead
```
saiskee, line 34: https://go.dev/blog/godoc

> To signal that an identifier should not be used, add a paragraph to its doc comment that begins with “Deprecated:” followed by some information about the deprecation.

saiskee:
```suggestion
// Deprecated: Deprecated due to typo. Use NewExponentialBackoff instead
```

saiskee:
Left a few comments, looks like this isn't a breaking change, but I'd tag @ilackarms or @EItanya for review since this is a core part of all our products.

saiskee (approved):

PR #477, opened 2022-04-25
CVE fix 0422

bewebi:
Bump golang.org/x/crypto dep to get fix for CVE-2022-27191


.github/workflows/pull_request.yaml
```diff
@@ -28,10 +28,10 @@ jobs:
       run: |
         wget https://github.com/aquasecurity/trivy/releases/download/v0.18.3/trivy_0.18.3_Linux-64bit.deb
         sudo dpkg -i trivy_0.18.3_Linux-64bit.deb
-    - name: Set up Go 1.14
+    - name: Set up Go 1.18
       uses: actions/setup-go@v1
       with:
-        go-version: 1.14
+        go-version: 1.18
```
jenshu, line 34: any reason not to use 1.18.1 here as well?

bewebi: Not particularly, just following the existing example of pinning minor version there and patch version in the code format action
Makes sense to keep them in sync one way or the other

bewebi: Looks like most of our other repos pin patch versions, so I'll go with that

jenshu (approved):

Issue #478, opened 2022-05-05
SecurityScanUtils should create fewer issues


bewebi:
The `CreateGithubIssuePerVersion` flag causes SecurityScanUtils to open an issue for every version of an image that contains a CVE, however this is overkill in many/most scenarios. When a patch for a given CVE on a given version is applied, it will usually be released as part of a _new_ version and the version in which the CVE was identified will contain the CVE in perpetuity.
Because of this, even if we close such security alert issues, they will always be reopened the next time the scan runs.

Therefore it would be beneficial to have a `CreateGithubIssuePerMinorVersion` flag that has a similar behavior, but only opens issues for the latest patch version of each minor version. This would ensure that we don't open an excessive number of redundant issues and ensures that once a version that patches a particular CVE is released we do not continue to open (or reopen) issues that alert about that CVE.

PR #479, opened 2022-05-09
securityscanutils: Only scan the latest patch version for a given minor release

nfuden:
Make sec gen less noisy and faster
BOT NOTES: 
resolves https://github.com/solo-io/go-utils/issues/478

### DIFF
--- a/changelog/v0.21.26/bump-crypto-dep.yaml
+++ b/changelog/v0.21.26/bump-crypto-dep.yaml
@@ -1,0 +1,4 @@
+changelog:
+  - type: NEW_FEATURE
+    issueLink: https://github.com/solo-io/go-utils/issues/478
+    description: Only create scans for the latest patches in a version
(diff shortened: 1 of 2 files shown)


nfuden:
Question: Do we feel that this should be gated with a flag? 
I think not as there is as far as I can tell no valid reason to be scanning old versions at least with how we handle cves but I defer to the reviewer.

jackstine:
> Question: Do we feel that this should be gated with a flag? I think not as there is as far as I can tell no valid reason to be scanning old versions at least with how we handle cves but I defer to the reviewer.

I think that having a flag would still defeat the purpose of the issue that was initially raised. I do not think we need a flag. 

The code LGTM

bewebi (commented):
IMO it would be more efficient and probably cleaner to sort `partialFilteredReleases`, then iterate through and add the first release for each minor version we see to `filteredReleases`

Separately, I'm wondering how/if you verified correct behavior? No need for automated tests, but a manual test at least would be good

nfuden:
I thought about doing that but then was annoyed with the major minor combo check. Ill throw it in.

I verified it via gloo with a go workspace. Ill do the same with the new format of using a slice.

nfuden:
Sooooo can has review?

bewebi (approved):
LGTM

PR #484, opened 2022-06-24
Remove context timeout

chunter0:
BOT NOTES: 
resolves https://github.com/solo-io/go-utils/issues/486


grpcutils/dial.go
```diff
@@ -59,8 +59,7 @@ func (o DialOpts) Dial(ctx context.Context) (*grpc.ClientConn, error) {
 		opts = append(opts, grpc.WithAuthority(o.Authority))
 	}
 	opts = append(opts, o.ExtraOptions...)
-	ctx, cancel := context.WithTimeout(ctx, time.Minute)
```
chunter0, line 62: Shouldn't the timeout be left to the user?

EItanya (commented):
@sam-heilbron @kdorosh is this code used in gloo today? I'm a little worried this could cause some pretty subtle bugs

nfuden:
Looks like its used in mesh currently but not in edge 

I agree that it should be configurable but in general dialing without timeout is not safe in a production use case so it seems safest that we just add an option and then set the default to be 1 min if not configured.

EItanya:
Agreed, I think an option makes sense, and if set to 0 it does nothing or something like that

grpcutils/dial.go
```diff
@@ -59,8 +59,7 @@ func (o DialOpts) Dial(ctx context.Context) (*grpc.ClientConn, error) {
...
-	defer cancel()
+
```
kdorosh, line 62: perhaps we should be returning cancel instead? plus the function signature change would make it obvious to devs about the change in behavior; so little risk of subtly breaking gloo edge @EItanya

chunter0: Does gloo edge crash when this function times out? @kdorosh

kdorosh: Per https://github.com/solo-io/go-utils/pull/484#issuecomment-1168645638 i don't think it's used

but i do think returning the cancel is better so that the caller has control over short-circuiting the timeout (which is also ideally configurable). just for better code; not because we would break gloo edge

chunter0: Actually I think it's easier to just create a context specific to the grpc connection and cancel it from the calling function. So maybe open an issue to upgrade and add this instead of modifying the signature? Wdyt?

chunter0: @nfuden @EItanya @kdorosh I don't feel strongly about the solution here. But looks like the census wants to allow the duration to be configurable and pass back a cancelFn to the caller?

EItanya: That'd be breaking,so if we're gonna break I'd rather just create a new function with no opinionated timeout that we can use in GM.

kdorosh: I'd argue its current behavior is confusing and borderline wrong, thus breaking has the added benefit of alerting users that behavior may not be what they expect

That said your approach Eitan is also very reasonable to me

chunter0: Hmm well if we want to keep the current behavior I can just close this PR no need to change it. I just noticed it while running the server with a debugger attached

EItanya: We're not saying always keep the same behavior, just have it be the default and allow overriding. But I also definitely see Kevin's point. I don't feel super strongly about it.

grpcutils/dial.go
```diff
@@ -59,8 +62,13 @@ func (o DialOpts) Dial(ctx context.Context) (*grpc.ClientConn, error) {
...
+	ctx, cancelFn := context.WithTimeout(ctx, (time.Minute * 1))
+	if o.Timeout > 0 {
+		ctx, cancelFn = context.WithTimeout(ctx, o.Timeout)
+	}
+	defer cancelFn()
```
kdorosh, line 70: why create a cancel and then only cancel it at the end of the function? why not just create a new context and let it go out of scope?

also i think this is slightly easier to read like so

```golang
        timeout := time.Minute
	if !o.Timeout.IsZero() {
		timeout = o.Timeout
	}
       ctx = context.WithTimeout(timeout)
```

grpcutils/dial.go
kdorosh: left a review

grpcutils/dial.go
chunter0: I think we agreed on this keeping the same behavior but making the timeout value optional.

grpcutils/dial.go
```diff
@@ -29,6 +29,9 @@ type DialOpts struct {
 
 	// additional options the caller wishes to inject
 	ExtraOptions []grpc.DialOption
+
+	// duration to wait for a connection to be established
```
kdorosh, line 33:
```suggestion
 	// duration to wait for a connection to be established. if zero, defaults to 1 minute
```

grpcutils/dial.go
chunter0: Could you suggest the changes? I don't see `IsZero` methood on time.Duration @kdorosh

kdorosh: updated, confused `time.Time` with `time.Duration` but we can still just compare it to zero

chunter0: Updated

PR #485, opened 2022-06-26
Trivy Scanning: Scan all images and only create issue for latest patch release

sam-heilbron:
# Description

- Fix bug in Trivy scanner which caused the scan to only run on the latest patch version
- Expose new flag that allows Github issues to be created only for the latest patch version, as opposed to all versions

# Context

## Fix Trivy scan bug
In https://github.com/solo-io/go-utils/issues/478, we identified that our historical scan implementation would create a Github issue for every version scanned that contained a vulnerability, and that this was extremely noisy to developers.

As a result, we added a slight adjustment to the code in https://github.com/solo-io/go-utils/pull/479. The unintended side effect of this was in addition to only creating github issues only for the latest patch release, we also only scanned the latest patch release. This speeds up our weekly job tremendously, but it also no longer produces results for older versions that users might be on. Since vulnerabilities might be identified after that fact, we want to make sure we're scanning all versions, since there might be a new vulnerability identified that we want to make users aware of.

## Expose new flag to only create issues for latest patch releases
We expose a new flag `CreateGithubIssueForLatestPatchVersion`, which when enabled will ensure that when a vulnerability is found for a given version, a github issue will be created only if that version is the latest patch release for that minor version.

## Move code into distinct files
Though these changes are somewhat noisy, and perhaps distract from the PR, I think they are valuable. Reading through the code, I found it challenging at times to distinguish the different components in our scanner. I copied tightly coupled code (trivy, github, predicates) into their own files so that the main scan file can have only the main scanning code.

## Stats logging follow-up
In https://github.com/solo-io/go-utils/pull/487 we added improvement to our stats utility, so that our logging endpoint handler does not throw a NPE. However, the goroutine responsible for waiting for a context to be cancelled and then initiating a server shutdown, was using the same context to shutdown the server, that it was using to determine if the server should be shutdown. This resulted in inconsistent behavior, and the server not being shutdown properly, which would lead to flakes.

This PR added a new context to be used for shutdown.

# Testing
Testing the security scanner end-to-end is challenging. A lot of the code relies on a `github.Client` that cannot be mocked. Therefore for this change I propose the following
1. I added unit tests to cover the case I am trying to solve, that we only create issues for the latest patch release
2. After merging this, I will upgrade the dependency in Gloo, and run the job and manually verify the results. 

## Manual Testing
I use https://github.com/solo-io/gloo/pull/6709 to manually verify these changes. In that PR, I outline the steps required to run the job locally and ways to verify both that (A) we scan all releases and (B) only generate github issues for the latest patch.

BOT NOTES: 
resolves https://github.com/solo-io/go-utils/issues/478

### DIFF
--- a/securityscanutils/github_writer.go
+++ b/securityscanutils/github_writer.go
@@ -1,0 +1,116 @@
+package securityscanutils
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/Masterminds/semver/v3"
+	"github.com/google/go-github/v32/github"
+	"github.com/rotisserie/eris"
+	"github.com/solo-io/go-utils/githubutils"
+)
+
+type GithubRepo struct {
+	RepoName string
+	Owner    string
+}
+
+func (r GithubRepo) Address() string {
+	return fmt.Sprintf("github.com/%s/%s", r.Owner, r.RepoName)
+}
+
+type GithubIssueWriter struct {
+	// The details about the Github repository
+	repo GithubRepo
+
+	// The client used to write the issues
+	client *github.Client
+
+	// The RepositoryReleasePredicate used to determine if a vulnerability
+	// associated with a certain release should be uploaded to GitHub
+	createGithubIssuePredicate githubutils.RepositoryReleasePredicate
+
+	// A local cache of all existing GitHub issues
+	// Used to ensure that we are updating existing issues that were created by previous scans
+	allGithubIssues []*github.Issue
+}
+
+func NewGithubIssueWriter(repo GithubRepo, client *github.Client, issuePredicate githubutils.RepositoryReleasePredicate) *GithubIssueWriter {
+	return &GithubIssueWriter{
+		repo:                       repo,
+		client:                     client,
+		createGithubIssuePredicate: issuePredicate,
+		allGithubIssues:            nil, // initially nil, we'll lazy load these
+	}
+}
+
+// Labels that are applied to github issues that security scan generates
+var labels = []string{"trivy", "vulnerability"}
+
+func (g *GithubIssueWriter) getAllGithubIssues(ctx context.Context) ([]*github.Issue, error) {
+	if g.allGithubIssues != nil {
+		return g.allGithubIssues, nil
+	}
+
+	issues, err := githubutils.GetAllIssues(ctx, g.client, g.repo.Owner, g.repo.RepoName, &github.IssueListByRepoOptions{
+		State:  "open",
+		Labels: labels,
+	})
+	if err != nil {
+		return nil, eris.Wrapf(err, "error fetching all issues from %s", g.repo.Address())
+	}
+	g.allGithubIssues = issues
+	return g.allGithubIssues, nil
+}
+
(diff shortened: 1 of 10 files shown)


nfuden:
As far as I can tell this is mainly addressing adding a repofilter that is applied only on the issue creation step which is a pretty easy update. Am I missing something else?

sam-heilbron:
> As far as I can tell this is mainly addressing adding a repofilter that is applied only on the issue creation step which is a pretty easy update. Am I missing something else?

Yup! That's it. I thought that was the set of changes we were trying to support

securityscanutils/github_writer.go
```diff
@@ -0,0 +1,127 @@
...
+var labels = []string{"trivy", "vulnerability"}
+
+// getAllGithubIssues returns the set of open issues in a Github repository that contain the trivy labels
+func (g *GithubIssueWriter) getAllGithubIssues(ctx context.Context) ([]*github.Issue, error) {
+	if g.allGithubIssues != nil {
+		// Maintain a local cache of issue to avoid re-requesting each time
+		return g.allGithubIssues, nil
+	}
+
+	issues, err := githubutils.GetAllIssues(ctx, g.client, g.repo.Owner, g.repo.RepoName, &github.IssueListByRepoOptions{
```
nfuden, line 61: This is dangerous as we shouldn't get all issues each time that we run for a given release as this is how we will get rate limited by github

sam-heilbron: Actually looking further. Any error that we encounter when creating a github issue will bubble up and cause the job to fail. This is what the previous behavior was as well. 
I'm more than happy to improve this if we think it's valuable

nfuden: Oh nice I see your intermediate has a check

securityscanutils/github_writer.go
```diff
@@ -0,0 +1,127 @@
...
+	if !g.shouldWriteIssue(release) {
+		// The GithubIssueWriter can be configured to only write issues for certain releases
+		return nil
+	}
+
+	// We can swallow the error here, any releases with improper tag names
+	// will not be included in the filtered list
+	versionToScan, _ := semver.NewVersion(release.GetTagName())
+
+	issueTitle := fmt.Sprintf("Security Alert: %s", versionToScan.String())
```
nfuden, line 91: nit: While we are changing things can we create our custom stringer
```suggestion
	issueTitle := fmt.Sprintf("Security Alert: %d.%d", versionToScan.Major(), versionToScan.Minor())
```

sam-heilbron: Spoke offline:

The challenge is that since we support the feature where you can create an issue per version, its possible to have multiple issues with the same major.minor and we'd therefore have issues with overlapping names.

By keeping the major.minor.patch in the title, we ensure that we create distinct issues. Also, this allows us to close issues when we resolve them, since a new issue (with a new title) will be created for the next version with a vulnerability.

nfuden (approved):
Yep this seems to cover it nicely

PR #487, opened 2022-07-06
Fix NPE in Stats Server Logging Handler

sam-heilbron:
# Description

Ensure that the stats server can handle requests to the `/logging` endpoint

# Context

## Gloo Edge Context
The logging endpoint is a useful feature for debugging components. In fact, we even recommend this feature in our production debugging docs for Gloo Edge: https://docs.solo.io/gloo-edge/latest/operations/debugging_gloo/#changing-logging-levels-and-more

## Problem
This feature has been broken for a while. You can see the [history](https://github.com/solo-io/go-utils/commits/master/stats/stats.go), but it has changed enough and been broken in different ways, that I decided to add a robust set of tests around the functionality.

In its most recent state, this handler is nil every time, unless the StartupOptions define a custom LogLevel.

## Solution
I introduced a context aspect to this server, to ensure that we can shutdown the server when the parent context is cancelled. Without this functionality, consecutive tests would fail, because the listener would never be shutdown and we would return stale config.

# How is this tested
- I added a variety of unit tests to ensure that we can use the logging endpoint to return the current log level, given a variety of startup options.
- I added a test to ensure that we can use the logging endpoint to change the log level, without requiring a restart to the server

I encountered some flakes, so to confirm that these tests don't introduce any flakes, I ran:
```
ginkgo -r -failFast -trace -progress -compilers=4 -failOnPending -noColor -randomizeSuites   -randomizeAllSpecs -untilItFails stats
```
and let it run for a while, before quitting

### DIFF
--- a/stats/stats.go
+++ b/stats/stats.go
@@ -58,7 +58,7 @@ func StartStatsServerWithPort(startupOpts StartupOptions, addhandlers ...func(mu
 	StartCancellableStatsServerWithPort(context.Background(), startupOpts, addhandlers...)
 }
 
-func StartCancellableStatsServerWithPort(ctx context.Context, startupOpts StartupOptions, addhandlers ...func(mux *http.ServeMux, profiles map[string]string)) {
+func StartCancellableStatsServerWithPort(ctx context.Context, startupOpts StartupOptions, customAddHandlers ...func(mux *http.ServeMux, profiles map[string]string)) {
 	// if the env var was provided (i.e., startup is conditional) and the value of that env var is not the expected value, then return and do nothing
 	if startupOpts.EnvVar != "" && os.Getenv(startupOpts.EnvVar) != startupOpts.EnabledValue {
 		return
@@ -68,48 +68,55 @@ func StartCancellableStatsServerWithPort(ctx context.Context, startupOpts Startu
 		contextutils.SetLogLevelFromString(envLogLevel)
 	}
 
+	go RunCancellableGoroutineStat(ctx)
+
 	// The running instance of the Stats server
 	var server *http.Server
 
-	go RunGoroutineStat()
+	addHandlers := append(customAddHandlers, addPprof, addStats)
+	go runGoroutineServerListenAndServe(ctx, startupOpts, server, addHandlers...)
+	go runGoroutineShutdownServer(ctx, server)
+}
 
-	go func() {
-		mux := new(http.ServeMux)
+func runGoroutineServerListenAndServe(
+	ctx context.Context,
+	startupOpts StartupOptions,
+	server *http.Server,
+	addHandlers ...func(mux *http.ServeMux, profiles map[string]string),
+) {
+	mux := new(http.ServeMux)
 
-		mux.Handle("/logging", getLoggingHandler(startupOpts))
+	mux.Handle("/logging", getLoggingHandler(startupOpts))
 
-		addhandlers = append(addhandlers, addPprof, addStats)
+	for _, addHandler := range addHandlers {
+		addHandler(mux, profileDescriptions)
+	}
 
-		for _, addhandler := range addhandlers {
-			addhandler(mux, profileDescriptions)
-		}
+	// add the index
+	mux.HandleFunc("/", Index)
 
-		// add the index
-		mux.HandleFunc("/", Index)
+	server = &http.Server{
+		Addr:    fmt.Sprintf(":%d", startupOpts.Port),
+		Handler: mux,
+	}
+	err := server.ListenAndServe()
+	if err == http.ErrServerClosed {
+		contextutils.LoggerFrom(ctx).Infof("Stats server closed")
+	} else {
+		contextutils.LoggerFrom(ctx).Warnf("Stats server closed with unexpected error: %v", err)
+	}
+}
 
-		server = &http.Server{
-			Addr:    fmt.Sprintf(":%d", startupOpts.Port),
-			Handler: mux,
-		}
-		err := server.ListenAndServe()
-		if err == http.ErrServerClosed {
-			contextutils.LoggerFrom(ctx).Infof("Stats server closed")
-		} else {
-			contextutils.LoggerFrom(ctx).Warnf("Stats server closed with unexpected error: %v", err)
-		}
-	}()
-
-	go func(parentContext context.Context) {
-		select {
-		case <-parentContext.Done():
-			if server != nil {
-				if err := server.Shutdown(parentContext); err != nil {
-					contextutils.LoggerFrom(parentContext).Warnf("Stats server shutdown returned error: %v", err)
-				}
-
+func runGoroutineShutdownServer(ctx context.Context, server *http.Server) {
+	select {
+	case <-ctx.Done():
+		if server != nil {
+			if err := server.Shutdown(ctx); err != nil {
+				contextutils.LoggerFrom(ctx).Warnf("Stats server shutdown returned error: %v", err)
 			}
+
 		}
-	}(ctx)
+	}
 }
 
 func getLoggingHandler(startupOpts StartupOptions) zap.AtomicLevel {
--- a/stats/stats.go
+++ b/stats/stats.go
@@ -103,10 +103,9 @@ func StartCancellableStatsServerWithPort(ctx context.Context, startupOpts Startu
 	// Run a separate goroutine to handle the server shutdown when the context is cancelled
 	go func() {
 		<-ctx.Done()
-			if server != nil {
-				if err := server.Shutdown(ctx); err != nil {
-					contextutils.LoggerFrom(ctx).Warnf("Stats server shutdown returned error: %v", err)
-				}
+		if server != nil {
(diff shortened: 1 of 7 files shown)


stats/stats.go
```diff
@@ -55,52 +55,61 @@ func ConditionallyStartStatsServer(addhandlers ...func(mux *http.ServeMux, profi
...
-			setLevel = zapcore.InfoLevel
-		}
+	if envLogLevel := os.Getenv(contextutils.LogLevelEnvName); envLogLevel != "" {
+		contextutils.SetLogLevelFromEnv(envLogLevel)
+	}
 
-		contextutils.SetLogLevel(setLevel)
+	mux := new(http.ServeMux)
 
-	} else if startupOpts.LogLevel != nil {
```
elcasteel, line 83: Why did you remove this `else`?

stats/stats.go
```diff
@@ -55,52 +55,61 @@ func ConditionallyStartStatsServer(addhandlers ...func(mux *http.ServeMux, profi
...
-			setLevel = zapcore.FatalLevel
-		default:
-			setLevel = zapcore.InfoLevel
-		}
+	if envLogLevel := os.Getenv(contextutils.LogLevelEnvName); envLogLevel != "" {
+		contextutils.SetLogLevelFromEnv(envLogLevel)
+	}
 
-		contextutils.SetLogLevel(setLevel)
+	mux := new(http.ServeMux)
```
ben-taussig-solo, line 71: You touched on it a little in slack, but could you clarify why we're moving this mux setup out of the goroutine?

stats/stats.go
elcasteel: Nevermind I figured out where it moved

sam-heilbron: Yeah I found the code a little confusing because we set the level in some places and get it from others. To try to simplify, I tried to separate those two

stats/stats.go
sam-heilbron: I wanted to be able to have another goroutine that could shutdown the server. I ran into the issue when writing tests that a single test would pass, but running multiple would fail, because the original goroutine was leaking / still running.

i'm going to move this construction back to the goroutine, back to how it used to be

ben-taussig-solo (commented):
Looks good to me. Waiting on tests + Ethan's review before approval.

stats/stats.go
```diff
@@ -99,8 +86,32 @@ func StartStatsServerWithPort(startupOpts StartupOptions, addhandlers ...func(mu
...
 	}()
+
+	go func(cancelContext context.Context) {
+		select {
+		case <-cancelContext.Done():
+			if server != nil {
+				_ = server.Shutdown(cancelContext)
+			}
+		}
+	}(ctx)
```
jackstine, line 104: LGTM, just saw that this was changed.

jackstine:
LGTM, I'll let Ethan approve it

stats/stats.go
```diff
@@ -99,8 +86,32 @@ func StartStatsServerWithPort(startupOpts StartupOptions, addhandlers ...func(mu
 
 		// add the index
 		mux.HandleFunc("/", Index)
-		http.ListenAndServe(fmt.Sprintf(":%d", startupOpts.Port), mux)
+
+		server = &http.Server{
+			Addr:    fmt.Sprintf(":%d", startupOpts.Port),
+			Handler: mux,
+		}
+		_ = server.ListenAndServe()
```
kdorosh, line 94: unhandled error

stats/stats.go
```diff
@@ -99,8 +86,32 @@ func StartStatsServerWithPort(startupOpts StartupOptions, addhandlers ...func(mu
...
 	}()
+
+	go func(cancelContext context.Context) {
+		select {
+		case <-cancelContext.Done():
+			if server != nil {
+				_ = server.Shutdown(cancelContext)
```
kdorosh, line 101: unhandled error. at least log?

stats/stats.go
sam-heilbron: Good call! It wasn't before and I lazily (incorrectly) just maintained. Added in a log line that should make things clearer

stats/stats_suite_test.go
```diff
@@ -22,6 +22,11 @@ func TestStats(t *testing.T) {
...
 
 var _ = BeforeSuite(func() {
-	// Tests in this suite expect the log level to be INFO to start
-	contextutils.SetLogLevel(zapcore.InfoLevel)
+	err := os.Setenv(stats.DefaultEnvVar, stats.DefaultEnabledValue)
+	Expect(err).NotTo(HaveOccurred())
+})
+
+var _ = AfterSuite(func() {
+	err := os.Setenv(stats.DefaultEnvVar, "")
```
kdorosh, line 30: nit, prefer `os.Unsetenv`

stats/stats_test.go
```diff
@@ -16,25 +19,37 @@ import (
...
-			resetEnvVars func()
 		)
 
 		BeforeEach(func() {
 			ctx, cancel = context.WithCancel(context.Background())
+
+			// Tests in this suite expect the log level to be INFO to start
+			contextutils.SetLogLevel(zapcore.InfoLevel)
+
+			err := os.Setenv(contextutils.LogLevelEnvName, "")
```
kdorosh, line 47: nit, prefer `os.Unsetenv()`

stats/stats.go
```diff
@@ -55,41 +55,28 @@ func ConditionallyStartStatsServer(addhandlers ...func(mux *http.ServeMux, profi
...
-	} else if startupOpts.LogLevel != nil {
-		logLevel = *startupOpts.LogLevel
+	if envLogLevel := os.Getenv(contextutils.LogLevelEnvName); envLogLevel != "" {
+		contextutils.SetLogLevelFromString(envLogLevel)
 	}
 
+	// The running instance of the Stats server
+	var server *http.Server
+
 	go RunGoroutineStat()
```
kdorosh, line 74: this leaks a goroutine itself, if we cancel the context this goroutine does not terminate

stats/stats.go
```diff
@@ -55,41 +55,28 @@ func ConditionallyStartStatsServer(addhandlers ...func(mux *http.ServeMux, profi
...
 
 	go func() {
 		mux := new(http.ServeMux)
 
-		mux.Handle("/logging", logLevel)
+		mux.Handle("/logging", getLoggingHandler(startupOpts))
 
 		addhandlers = append(addhandlers, addPprof, addStats)
```
kdorosh, line 81: this seems like a race too.. what if the caller depended on the `addhandlers` array behind the slice copy passed in? I think we should copy the array contents before using them in the goroutine

stats/stats.go
```diff
@@ -55,52 +55,72 @@ func ConditionallyStartStatsServer(addhandlers ...func(mux *http.ServeMux, profi
...
+			contextutils.LoggerFrom(ctx).Infof("Stats server closed")
+		} else {
+			contextutils.LoggerFrom(ctx).Warnf("Stats server closed with unexpected error: %v", err)
+		}
 	}()
+
+	// Run a separate goroutine to handle the server shutdown when the context is cancelled
+	go func() {
+		select {
+		case <-ctx.Done():
```
kdorosh, line 106: nit, no need to select with a single case. we can just do `<-ctx.Done()`
```suggestion
		<-ctx.Done()
````

**Review by kdorosh — COMMENTED** [review]
lgtm, leaving for others to approve

File: stats/stats.go (hunk 9 above)
Comment (sam-heilbron) [reply stats/stats.go]: great catch here, updated

File: stats/stats.go (hunk 8 above)
Comment (sam-heilbron) [reply stats/stats.go]: yeah good call, updated

**Review by elcasteel — APPROVED** [review]

File: stats/stats_test.go [hunk 11]
```diff
@@ -0,0 +1,168 @@
...
+			// that we wait for the port to be released, so we can start up a next server on the subsequent test
+			EventuallyPortAvailable(startupOptions.Port)
+		})
+
+		When("StartOptions are default", func() {
+
+			BeforeEach(func() {
+				startupOptions = stats.DefaultStartupOptions()
+
+				stats.StartCancellableStatsServerWithPort(ctx, startupOptions)
```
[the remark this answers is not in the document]
sam-heilbron: There should only be 1 beforeEach per test that call this. The reason for doing it was that each set of tests require different startup options to run with. I figured my options were:
1. Define the startupOptions in the BeforeEach of each test group, and call StartStatsServer in a JustBeforeEach at the outer Context
2. Call StartStatsServer in each of the individual `It` and letting the startupOptions be constructed in the BeforeEach
3. This approach


I'm not strongly bound to this and am certainly open to an alternative approach if you have one in mind that you think would be easier to reason through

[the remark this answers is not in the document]
elcasteel: Ah, the narrow github diff makes reading indentation levels harder and I was _convinced_ that the second call was nested in this first one. 
This approach seems fine to me. I don't think there's a a great way to handle the way ginkgo tests end up very deeply nested.

PR #488, opened 2022-07-11
Improve Scanning Performance

sam-heilbron:
# Describe
Our Trivy scans had a small bug wherein a scan would run and no matter what the result of the scan was, the scan would be rerun with an increasing delay causing each scan to take upwards of three minutes to run where a manual scan would have taken milliseconds/seconds, this caused our scans on multiple images each with multiple tagged versions to take upwards of six hours. We solved this by adding exit code checking logic in the `executeScanWithRetries()` function.

BOT NOTES: 
resolves https://github.com/solo-io/go-utils/issues/490


securityscanutils/trivy_scanner_test.go
```diff
@@ -0,0 +1,83 @@
...
+
+		//Error occurs when all trivy scan arguments are empty
+		Expect(err).To(HaveOccurred())
+		Expect(completed).To(Equal(false))
+		Expect(vulnFound).To(Equal(false))
+
+	})
+
+	FContext("Benchmark", func() {
+		It("Should do repeated scans efficiently", func() {
```
bewebi, line 71: It looks like we're scanning three different images
Why do we consider the scans to be repeated and why do we expect them to take <4s?

Bslabe123: Dont we want to scan multiple images or the same image multiple times? Also do we have a time in mind for how quickly we want these scans to finish? I made a guess based off of (local testing time) * 2.

bewebi: > Dont we want to scan multiple images or the same image multiple times?

I think multiple images is what we're doing both in this test and in practice, but the test name is misleading, implying the same image is scanned multiple times

> Also do we have a time in mind for how quickly we want these scans to finish?

This is sort of what I was asking myself. I'm not sure how much time it should take or what our threshold should be specifically. 
If the previous implementation was an order of magnitude more and we intend to verify that we have reduced time by that order of magnitude we should be able to set a threshold at the lower order of magnitude and conservative enough to be hit consistently. 
A comment indicating what it would have taken previously and what it should take now that we avoid retries might be helpful context

securityscanutils/trivy_scanner_test.go
```diff
@@ -0,0 +1,81 @@
...
+	"path/filepath"
+	"time"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/solo-io/go-utils/osutils/executils"
+	. "github.com/solo-io/go-utils/securityscanutils"
+)
+
+var _ = Describe("Trivy Scanner", func() {
```
sam-heilbron, line 16: I like these end to end tests that validate that we can execute trivy directly. I think it could be valuable as well (and the main reason I made the CmdExecutor injectable) to run the trivy scanner with a mock command executor, and confirm that when we return certain status codes and errors, that the trivy scanner only retries a certain number of times. This way we can validate how our TrivyScanner responds to the cmd output.

securityscanutils/trivy_scanner.go
```diff
@@ -31,73 +41,71 @@ func RunTrivyScan(image, version, templateFile, output string) (bool, bool, erro
...
+		attemptStart := time.Now()
+		if t.executeCommand == nil {
+			return false, false, eris.Errorf("TrivyScanner has no executeCommand function defined")
+		}
+		out, statusCode, err = t.executeCommand(trivyScanCmd)
+		logger.Debugf("Trivy returned %d after %s", statusCode, time.Since(attemptStart).String())
 
-		// If there is no error, don't retry
+		// If we receive the expected status code, the scan completed, don't retry
+		if statusCode == VulnerabilityFoundStatusCode {
```
sam-heilbron, line 81: Given this new condition (and all exit conditions), is this true for all versions of trivy? Do different version of trivy have different behaviors?

securityscanutils/trivy_scanner.go
```diff
@@ -31,73 +41,71 @@ func RunTrivyScan(image, version, templateFile, output string) (bool, bool, erro
...
 		err        error
 	)
 
-	for attempt := 0; attempt < retryCount; attempt++ {
-		trivyScanCmd := exec.Command("trivy", trivyScanArgs...)
-		out, statusCode, err = executils.CombinedOutputWithStatus(trivyScanCmd)
+	for attempt := 0; attempt < t.scanMaxRetries; attempt++ {
+		trivyScanCmd := exec.Command("trivy", scanArgs...)
+		attemptStart := time.Now()
+		if t.executeCommand == nil {
```
sam-heilbron, line 74: I'd be ok removing this guard. When you instantiate a TrivyScanner you pass the executeCommand value. To me, if you pass nil and get an NPE that would be acceptable.

securityscanutils/trivy_scanner.go
```diff
@@ -31,73 +41,71 @@ func RunTrivyScan(image, version, templateFile, output string) (bool, bool, erro
...
+			return false, false, eris.Errorf("TrivyScanner has no executeCommand function defined")
+		}
+		out, statusCode, err = t.executeCommand(trivyScanCmd)
+		logger.Debugf("Trivy returned %d after %s", statusCode, time.Since(attemptStart).String())
```
sam-heilbron, line 78: I like this log line! I wonder, now that this will occur many times, whether the other log messages should be at the info level?

Is there any concern with flooding the logs?

securityscanutils/trivy_scanner_test.go
```diff
@@ -0,0 +1,81 @@
...
+	Context("Benchmark", func() {
+		It("Should do repeated scans efficiently", func() {
+			inputImage = "quay.io/solo-io/gloo:1.11.1"
+			attemptStart := time.Now()
+			for i := 0; i < 10; i++ {
+				_, _, err := t.ScanImage(context.TODO(), inputImage, inputMarkdownTemplateFile, outputFile)
+				Expect(err).NotTo(HaveOccurred())
+			}
+			attemptEnd := time.Since(attemptStart)
+			Expect(attemptEnd).To(BeNumerically("<", 20*time.Second))
```
sam-heilbron, line 78: I think we should run this test locally many times, gather some benchmarks, and then use something like https://github.com/solo-io/ext-auth-service/blob/26052235c2f91e9d0a0f77a340c636cd3aa2b324/pkg/config/opa/config_test.go#L232 to have some benchmarks

securityscanutils/securityscan.go
```diff
@@ -137,7 +142,7 @@ func (s *SecurityScanner) GenerateSecurityScans(ctx context.Context) error {
 
 			// Only generate sarif files if we are uploading code scan results to github
 			if repo.Opts.UploadCodeScanToGithub {
-				err = repo.RunGithubSarifScan(release, sarifTplFile)
+				err = repo.RunGithubSarifScan(ctx, release, sarifTplFile)
```
sam-heilbron, line 145: Since this is an exported function, this is technically a breaking change. You could argue that this function should never have been exported in the first place as its intended to be called internally. i'll leave it up to you whether you want to make this a minor version bump as a result, or just acll it out loudly in the PR.

As other teams pull in these changes, we need to make it easy to identify how to upgrade to this version and what the impacts are

securityscanutils/trivy_scanner.go
```diff
@@ -31,73 +41,71 @@ func RunTrivyScan(image, version, templateFile, output string) (bool, bool, erro
...
+			// even if some releases failed and we didn't publish images for those releases
+			// this error used to happen if a release was a pre-release and therefore images
+			// weren't pushed to the container registry.
+			// we have since filtered out non-release images from being scanned so this warning
+			// shouldn't occur, but leaving here in case there was another edge case we missed
+			return false, false, nil
 		}
 
-		backoffStrategy(attempt)
+		t.scanBackoffStrategy(attempt)
```
sam-heilbron, line 104: What error could occur that results in a retry? What are we protecting aginst?

securityscanutils/securityscan.go
```diff
@@ -41,6 +44,8 @@ type SecurityScanRepo struct {
 	// should be run through our scanner
 	scanReleasePredicate githubutils.RepositoryReleasePredicate
 
+	trivyScanner *TrivyScanner
```
sam-heilbron, line 47: Now that we define a trivy scanner, could we mock it out and have tests that validate our whole job? This way we can validate the job without having to interact with trivy directly? Perhaps in a separate PR also

sam-heilbron (commented):
Looks good!

I think we can further harden these tests to avoid hitting the bug we encountered.

I also thinks its valuable to check out the mesh team and portal repo to see if the retry bug we experience in gloo also hits them. If so, we can be safe, but if not, there may be a further behavior that we haven't accounted for.

securityscanutils/trivy_scanner.go
Bslabe123: Cleaned up the log statements to prevent flooding

securityscanutils/trivy_scanner.go
Bslabe123: as long as the --exit-code flag is set (which it is in our code), then finding vulnerabilities will always result in an exit code specified by this flag, this is true for all versions that support the --exit-code flag

securityscanutils/securityscan.go
Bslabe123: Might want to do a minor version bump just to keep consistent

securityscanutils/trivy_scanner_test.go
sam-heilbron: Let''s use https://github.com/solo-io/ext-auth-service/blob/master/pkg/config/opa/config_test.go#L232 to introduce some more targeted benchmark testing

sam-heilbron (commented):
Looking good!

A few more thoughts around ways to really harden our testing.

It would be great too if there were a test that failed before these changes and passes only with them.

Any luck reaching out to other teams that use this library to see if they experience latency when scanning? We should also be able to run those locally to confirm

securityscanutils/trivy_scanner_test.go
```diff
@@ -0,0 +1,108 @@
...
+		tMock := NewTrivyScanner(MockCmdExecutor)
+		completed, vulnFound, err := tMock.ScanImage(context.TODO(), inputImage, inputMarkdownTemplateFile, outputFile)
+
+		Expect(err).To(BeNil())
+		Expect(completed).To(Equal(true))
+		Expect(vulnFound).To(Equal(false))
+	})
+
+	It("Times out while backing off and retrying when mock executor returns error", func() {
+		MockCmdExecutor := func(cmd *exec.Cmd) ([]byte, int, error) {
```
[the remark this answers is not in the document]
Bslabe123: This test doesn't look at retries so the description may be narrow here, but it makes sure that ScanImage returns an error if the MockCmdExecutor returns an error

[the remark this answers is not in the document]
Bslabe123: Should I look at modifying the behavior or the description?

securityscanutils/trivy_scanner_test.go
```diff
@@ -0,0 +1,108 @@
...
+		//Error occurs when all trivy scan arguments are empty
+		Expect(err).To(HaveOccurred())
+		Expect(completed).To(Equal(false))
+		Expect(vulnFound).To(Equal(false))
+
+	})
+
+	It("Returns the correct status code with mock executor returning no errors", func() {
+		MockCmdExecutor := func(cmd *exec.Cmd) ([]byte, int, error) {
+			return nil, 12345, nil
```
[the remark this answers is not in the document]
Bslabe123: In this situation there would be no retries because no error is returned from the executor, if the executor returns no error the trivy scan will exit out there regardless of what the executor returned for the other outputs, and it only returns true for vulnFound if the exit code equals the one we specify, otherwise its always false.

securityscanutils/trivy_scanner_test.go
```diff
@@ -0,0 +1,108 @@
...
+		inputImage = "quay.io/solo-io/gloo:1.11.13245"
+		completed, vulnFound, err := t.ScanImage(context.TODO(), inputImage, inputMarkdownTemplateFile, outputFile)
+
+		Expect(err).NotTo(HaveOccurred())
+		Expect(completed).To(Equal(false))
+		Expect(vulnFound).To(Equal(false))
+
+	})
+
+	It("Returns error from Exec via Timeout", func() {
```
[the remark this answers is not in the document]
Bslabe123: This is testing what happens when the executor returns an error that doesn't correspond to an image not found error and defaults out to a backoff and retry, in this case it should retry indefinitely and time out. I used empty inputs to get an invalid input error on the executor.

securityscanutils/trivy_scanner.go
```diff
@@ -31,73 +41,67 @@ func RunTrivyScan(image, version, templateFile, output string) (bool, bool, erro
...
-		// even if some releases failed and we didn't publish images for those releases
-		// this error used to happen if a release was a pre-release and therefore images
-		// weren't pushed to the container registry.
-		// we have since filtered out non-release images from being scanned so this warning
-		// shouldn't occur, but leaving here in case there was another edge case we missed
-		if IsImageNotFoundErr(string(out)) {
-			log.Warnf("image %s not found for version %s", image, version)
-			return false, false, nil
-		}
-		return false, false, eris.Wrapf(err, "error running trivy scan on image %s, version %s, Logs: \n%s", image, version, string(out))
```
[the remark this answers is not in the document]
Bslabe123: We couldnt rely on other CI tests to check that?

securityscanutils/trivy_scanner_test.go
```diff
@@ -0,0 +1,109 @@
...
+	)
+
+	JustBeforeEach(func() {
+		t = NewTrivyScanner(executils.CombinedOutputWithStatus)
+		inputMarkdownTemplateFile, err = GetTemplateFile(MarkdownTrivyTemplate)
+		Expect(err).NotTo(HaveOccurred())
+		outputDir, err := ioutil.TempDir("", "")
+		Expect(err).NotTo(HaveOccurred())
+		outputFile = filepath.Join(outputDir, "test_report.docgen")
+		inputImage = "quay.io/solo-io/gloo:1.11.1"
```
bewebi, line 35: It would be ideal to write tests that don't depend on the ability to reach our Docker repo as that is liable to flake
Mocking the command executor would be one solution

securityscanutils/trivy_scanner_test.go
```diff
@@ -0,0 +1,109 @@
...
+	})
+
+	JustAfterEach(func() {
+		err := os.RemoveAll(outputDir)
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	It("Finds vulnerabilities", func() {
+		completed, vulnFound, err := t.ScanImage(context.TODO(), inputImage, inputMarkdownTemplateFile, outputFile)
```
bewebi, line 44: It's implicit that the provided image has vulnerabilities, but we should make it explicit with a comment

securityscanutils/trivy_scanner_test.go
```diff
@@ -0,0 +1,109 @@
...
+		Expect(completed).To(Equal(false))
+		Expect(vulnFound).To(Equal(false))
+	})
+
+	Context("Benchmark", func() {
+		It("Should do repeated scans efficiently", func() {
+			inputImage = "quay.io/solo-io/gloo:1.11.1"
+			attemptStart := time.Now()
+			samples := 25
+			//for reference: when testing locally these samples were all between 535.581875ms and 879.919541ms
```
bewebi, line 100: Adding context around what it was previously would also be helpful

bewebi (commented):
Some minor comments, also agree with Sam's feedback

securityscanutils/trivy_scanner.go
```diff
@@ -1,28 +1,38 @@
...
-		return false, false, eris.Wrap(err, "trivy is not on PATH, make sure that the trivy v0.18 is installed and on PATH")
+type CmdExecutor func(cmd *exec.Cmd) ([]byte, int, error)
+
+type TrivyScanner struct {
+	executeCommand      CmdExecutor
+	scanBackoffStrategy func(int)
+	scanMaxRetries      int
+}
+
+func NewTrivyScanner(executeCommand CmdExecutor) *TrivyScanner {
```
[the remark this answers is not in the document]
Bslabe123: We hardcoded these values to keep consistency with the hard coded constants pre-refactoring. Its totally possible to allow them to be passed though.

securityscanutils/trivy_scanner_test.go
Bslabe123: Fixed! I would assume the benchmarking tests have to hit the Docker repo though right?

securityscanutils/trivy_scanner_test.go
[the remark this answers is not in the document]
sam-heilbron: No that's fair! I guess my main point is, we have some tests to validate that the ScanImage behavior works correctly and it would be helpful to know that it retries/doesn't retry correctly. Doesn't have to be this test in particular

securityscanutils/trivy_scanner.go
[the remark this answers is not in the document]
sam-heilbron: I don't believe we have that many CI tests in this space, though I'd double check to see what cases we cover

securityscanutils/trivy_scanner_test.go
sam-heilbron: I think this is a good point to raise and I see a couple of potential ideas around this:
1. I think it would be valuable to unit test the behavior by mocking the cmdExecutor so we're not hitting trivy directly (this was the intention of making it injectable)
2. I think it would also be valuable to have an integration test that does hit trivy and docker, since our method is designed to be resilient to flakes, since we don't want those to cause the job to fail. If the test is flaky that to me indicates that the job would be similarly flakey (which we want to avoid)

RE bencharking: I guess we don't need to benchmark this test specifically, but I do think we want to ensure that we are retrying at the proper times and not retrying at other times

securityscanutils/trivy_scanner_test.go
Bslabe123: Added! I included the ranges of response times from my local testing environment rather than each test individually, I can add back the multiple images test no problem as well if we need that, but as ben mentioned do we want to keep this test as trivy isn't something we have control over?

securityscanutils/trivy_scanner_test.go
Bslabe123: Added!

securityscanutils/trivy_scanner.go
Bslabe123: Removed!

securityscanutils/trivy_scanner_test.go
Bslabe123: I added in the range of response times I got from local testing, I benchmarked the sum of response times which might be better if we ever decide to multithread the trivy scans

securityscanutils/trivy_scanner.go
Bslabe123: Added!

securityscanutils/securityscan.go
Bslabe123: Any functions in particular we should make private other than this one?

securityscanutils/trivy_scanner_test.go
```diff
@@ -0,0 +1,108 @@
...
+	"path/filepath"
+	"time"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/solo-io/go-utils/osutils/executils"
+	. "github.com/solo-io/go-utils/securityscanutils"
+)
+
+var _ = FDescribe("Trivy Scanner", func() {
```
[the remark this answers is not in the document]
Bslabe123: Fixed!

securityscanutils/trivy_scanner_test.go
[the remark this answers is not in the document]
Bslabe123: Done!

securityscanutils/trivy_scanner_test.go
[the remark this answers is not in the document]
Bslabe123: Gotcha, I think we test other functionality in other tests thoroughly enough to make this a "retries correctly" test if that  sounds good with you

securityscanutils/trivy_scanner_test.go
Bslabe123: Added

securityscanutils/trivy_scanner_test.go
Bslabe123: Fixed this with a mock executor which always returns the error status code

securityscanutils/trivy_scanner_test.go
Bslabe123: That works out really well! The tests as they stand now are all run with mock executors so we can test our logic independent of trivy and we multipurpose our benchmarking test as an integration test as well since it has to hit trivy.

bewebi (commented):
Resolved most of the out standing conversations and left what I expect to be the last round of nits

securityscanutils/trivy_scanner_test.go
```diff
@@ -49,7 +52,9 @@ var _ = Describe("Trivy Scanner", func() {
 	})
 
 	It("Cannot find Image", func() {
-		inputImage = "quay.io/solo-io/gloo:1.11.13245"
+		t = NewTrivyScanner(func(cmd *exec.Cmd) ([]byte, int, error) {
+			return []byte("Error containing the string 'No such image: '"), VulnerabilityFoundStatusCode + 1, eris.Errorf("Unread error")
```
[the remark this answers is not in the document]
Bslabe123: Anything that isn't `VulnerabilityFoundStatusCode` would correspond to the image not being pulled, i just went with +1 to make it look realistic but this could in theory be anything.

securityscanutils/trivy_scanner.go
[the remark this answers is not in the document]
Bslabe123: We can test he executeScanWithRetries() function but its outputs are still tested. That new scanCompleted var is passed directly out so we have all of our checks for that output

securityscanutils/trivy_scanner_test.go
bewebi: Another way to do this that's a little cleaner/idiomatic might be to use a [`Consistently()`](https://onsi.github.io/gomega/#consistently) block, ensuring that we don't flake over a certain period of time rather than a set number of retries

bewebi: But if you look into that and don't think it's preferable I'll defer

sam-heilbron (commented):
This is abreaking change, can we update the changelog to call that out, and make this a minor version bump?

bewebi (approved):
LGTM

PR #493, opened 2022-08-02
Trivyignore

Bslabe123:
BOT NOTES: 
resolves https://github.com/solo-io/solo-projects/issues/3907


securityscanutils/securityscan.go
```diff
@@ -30,6 +30,9 @@ import (
 type SecurityScanner struct {
 	Repos        []*SecurityScanRepo
 	githubClient *github.Client
+
+	//trivyIgnore file contents
```
nfuden, line 34: sure but what does trivyignore do? Link or a quick comment?

securityscanutils/securityscan.go
```diff
@@ -30,6 +30,10 @@ import (
 type SecurityScanner struct {
 	Repos        []*SecurityScanRepo
 	githubClient *github.Client
+
+	// trivyignore file contents
+	// A trivyignore file contains a newline separated list of CVEs in the form CVE-20XX-XXXX to be ignored in trivy scans
+	TrivyIgnoreContents string
```
sam-heilbron, line 36: This is listed globally for the SecurityScanner..would it make sense to do this per repo since each repo may have a different set of CVEs it wants to ignore?

securityscanutils/securityscan.go
```diff
@@ -126,6 +130,14 @@ func (s *SecurityScanner) GenerateSecurityScans(ctx context.Context) error {
...
 	}()
 
+	//Recreate a .trivyignore file locally
+	ignoreFile, err := ioutil.TempFile("", "")
+	if err != nil {
+		return err
+	}
+	ignoreFile.WriteString(s.TrivyIgnoreContents)
+	defer os.Remove(ignoreFile.Name())
+
```
sam-heilbron, line 140: I think this leaks information about trivy scanning into the broader scanner. Ideally, we should isolate all trivy logic into the TrivyScanner struct. Instead of passing this file around, could we instantiate the trivyScanner with the list of cves to ignore, or call RunTrivyScan and pass the list of cves? That way we can keep the fact that it's a file (and where that file lives) hidden from the rest of the components.

securityscanutils/trivy_scanner_test.go
```diff
@@ -104,9 +104,21 @@ var _ = Describe("Trivy Scanner", func() {
 			inputImage = "quay.io/solo-io/gloo:1.11.1"
 			samples := 8
 			for i := 0; i < samples; i++ {
-				_, _, err := t.ScanImage(context.TODO(), inputImage, inputMarkdownTemplateFile, outputFile)
+				_, _, err := t.ScanImage(context.TODO(), inputImage, inputMarkdownTemplateFile, outputFile, "")
 				Expect(err).NotTo(HaveOccurred())
 			}
 		})
+
+		It("ignores CVEs in trivyIgnore", func() {
```
sam-heilbron, line 112: I think this test should first run a scan, and produce vulnerabilities. Then we can run another scan with the ignore file and demonstrate that with the file, those same vulnerabilities are not discovered.

What happens if 1.9.21 finds new vulnerabilities 1 month from now? Won't this test start failing because we will now find vulnerabilities? How do we want to address that?

Bslabe123:
Not needed, trivyignore works as intended

PR #495, opened 2022-08-04
return error if unable to retrieve ChanelogNotes

bewebi:


nfuden (approved):
Given that this is a flakey and hard to repro I am good merging this to have more descriptive errors

PR #506, opened 2023-02-03
Ginkgo v2

sam-heilbron:
# Description
Upgrade ginkgo library to use v2

# Context
Projects which consume go-utils in test code, require that go-utils depends on ginkgo-v2 before they can be upgraded

## Pull Request GHA Changes
I updated the job trigger to run on pull requests to `main`. My intention is to change the default branch in this repo from master to main after this merges.

Our tests don't actually use a kube cluster, so I removed the setup to create a cluster and install kubectl

Our action used an old base image. I updated that. I also updated our gcloud setup as that broke on newer ubuntu versions.

## Makefile
I added some make targets that will be easily consumable by devs. See `make help` for a list of useful targets:
```
help                                Output the self-documenting make targets
clean                               Clean any local assets
test                                Run tests in the {TEST_PKG}
test-with-coverage                  Run tests in the {TEST_PKG} with coverage
```

# Testing
- [x] Run `TEST_PKG=./hashutils make test`
- [x] Run `TEST_PKG=./hashutils make test-with-coverage`


jackstine:
I found a reference that do not use `v2`
[ginko/config](https://github.com/solo-io/go-utils/blob/ginkgo-v2/testutils/port/port.go#L7-L8)

.github/workflows/pull_request.yaml
```diff
@@ -45,9 +45,8 @@ jobs:
     - name: Run tests
       env:
         GITHUB_TOKEN: ${{ secrets.CLOUDBUILD_GITHUB_TOKEN }}
-      run: |
-        go install -v github.com/onsi/ginkgo/ginkgo@v1.16.5 && export PATH=$PATH:$(go env GOPATH)/bin/
-        ginkgo -r -failFast -randomizeSuites -randomizeAllSpecs -skipPackage=./installutils/kubeinstall,./debugutils/test
+        TEST_PKG: ./... # Run all tests
+      run: make test
     - uses: testspace-com/setup-testspace@v1
       with:
```
gunnar-solo, line 51: what goes on, here? Specifically: we now have some newly modified code coverage+junit report stuff. I've never actually seen this testspace server/data

I'm asking, as I think it might be more useful to store these as github actions artifacts

Makefile
```diff
@@ -13,4 +16,33 @@ install-go-tools:
...
+# Tests
+#----------------------------------------------------------------------------------
+
+GINKGO_VERSION ?= 2.5.0 # match our go.mod
+GINKGO_ENV ?= GOLANG_PROTOBUF_REGISTRATION_CONFLICT=ignore ACK_GINKGO_DEPRECATIONS=$(GINKGO_VERSION)
+GINKGO_FLAGS := -v -tags=purego -compilers=4 -fail-fast -randomize-suites -randomize-all -skip-package=./installutils/kubeinstall,./debugutils/test
+GINKGO_REPORT_FLAGS := --json-report=test-report.json --junit-report=junit.xml
+GINKGO_COVERAGE_FLAGS := --cover --covermode=count --coverprofile=coverage.cov
+
+# This is a way for a user executing `make test` to be able to provide flags which we do not include by default
```
bewebi, line 31: Is it plausible a user might want to override/opt into some of the default flags too?
ie should all of these have `?=`?

go.mod
```diff
@@ -15,7 +15,7 @@ require (
 	github.com/go-git/go-git/v5 v5.4.1
 	github.com/gogo/protobuf v1.3.1
 	github.com/golang/mock v1.5.0
-	github.com/golang/protobuf v1.4.3
+	github.com/golang/protobuf v1.5.2
```
bewebi, line 18: Was this necessary for the ginkgo bump? Is there a chance this will make things hairier when rolling out to other repos?

Makefile
```diff
@@ -13,4 +16,33 @@ install-go-tools:
...
+	goimports -w .
+
+#----------------------------------------------------------------------------------
+# Tests
+#----------------------------------------------------------------------------------
+
+GINKGO_VERSION ?= 2.5.0 # match our go.mod
+GINKGO_ENV ?= GOLANG_PROTOBUF_REGISTRATION_CONFLICT=ignore ACK_GINKGO_DEPRECATIONS=$(GINKGO_VERSION)
+GINKGO_FLAGS := -v -tags=purego -compilers=4 -fail-fast -randomize-suites -randomize-all -skip-package=./installutils/kubeinstall,./debugutils/test
+GINKGO_REPORT_FLAGS := --json-report=test-report.json --junit-report=junit.xml
```
bewebi, line 28: Would it be worthwhile to document in a README or some such that these reports are generated (and how)?
It may be opaque to developers especially now that it's fully hidden in a Makefile and not referenced in code

bewebi (commented):
A few questions, but looks good!

.github/workflows/pull_request.yaml
sam-heilbron: We used to execute a ginkgo command with several arguments in a github action. What I have done is migrate that command to a make target to mirror our other repos. This way, devs can execute the same command that is run in CI (`make test`). I updated the flags we provide to use the new v2 syntax but did not change it apart from that.

Makefile
sam-heilbron: Yeah that's fair. What do you think is the best way to approach this? This will only be generated if you run `make test-with-coverage`. One option would be to document that target, and make it visible in `make help`. But another would be to create a README. I'm happy to for either, what do you think?

Makefile
sam-heilbron: Yeah probably safest to make them all optional. I figure these are the flags that we always want to run, since they are run in CI, so I can't think of a good situtation where you wouldn't want these assigned. However, I'm up for more opt-in functionality so I like this proposal and will update

go.mod
sam-heilbron: I will explore. All of these deps were behind and I intentionally updated them to match what we have in our projects (GE, GME). I will make sure that we can pull in these changes successfully and if there are challenges will reconsider

bewebi: If it's bringing the dep up to date with our other repos that's great
I'd be nervous if it was pushing a dep ahead of other repos

Makefile
bewebi: Ah, I missed that part
Documenting `test-with-coverage` would be great, either in `make help` or README
I'd probably prefer README since it's the first place I'd look, but not a strongly held opinion

bewebi (dismissed):
LGTM

gunnar-solo (dismissed):

jackstine (approved):

gunnar-solo (approved):
make it so.

sam-heilbron:
We're getting rate-limited on our Github client so tests are failing. These utilities are important to communicate directly wit hthe API so I'm not going to build any workaround for them.

Makefile
```diff
@@ -13,4 +38,43 @@ install-go-tools:
...
+
+.PHONY: test
+test: install-test-tools ## Run tests in the {TEST_PKG}
+	$(GINKGO_ENV) ginkgo \
+	$(GINKGO_FLAGS) $(GINKGO_REPORT_FLAGS) $(GINKGO_USER_FLAGS) \
+	$(TEST_PKG)
+
+.PHONY: test-with-coverage
+test-with-coverage: GINKGO_FLAGS += $(GINKGO_COVERAGE_FLAGS) ## Run tests in the {TEST_PKG} with coverage
+test-with-coverage: test
```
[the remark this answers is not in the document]
sam-heilbron: Perhaps I'm misunderstanding but isn't that what we're doing? When running `make test-with-coverage` we run ginkgo with additional coverage flags (`-cover`). Can you link me to some other way to do this that you had in mind?

gunnar-solo (approved):

PR #507, opened 2023-02-07
versionutils: support more than 1 label variant and their strict order

nfuden:
The default list can be overridden by any consumer if they so choose but it states when a format is better than alphanumeric. 
One consideration reviewers should take into account is whether it makes sense to use string prefix or to convert these things into a regex. As it is now rcblah1 will be treated like rc1 which may not be ideal but in general it makes the code cleaner.
BOT NOTES: 
resolves https://github.com/solo-io/go-utils/issues/0000

### DIFF
--- a/changelogutils/changelog_test.go
+++ b/changelogutils/changelog_test.go
@@ -18,13 +18,13 @@ import (
 	"github.com/spf13/afero"
 )
 
-var _ = Describe("ChangelogTest", func() {
+var _ = FDescribe("ChangelogTest", func() {
 
 	var _ = Context("GetProposedTag", func() {
 		getProposedTag := func(latestTag, changelogDir, tag string) error {
 			fs := afero.NewOsFs()
 			actualTag, actualErr := changelogutils.GetProposedTag(fs, latestTag, changelogDir)
-			Expect(actualTag).To(BeEquivalentTo(tag))
+			Expect(actualTag).To(BeEquivalentTo(tag), actualErr)
 			return actualErr
 		}
 
--- a/changelogutils/changelog_test.go
+++ b/changelogutils/changelog_test.go
@@ -18,7 +18,7 @@ import (
 	"github.com/spf13/afero"
 )
 
-var _ = FDescribe("ChangelogTest", func() {
+var _ = Describe("ChangelogTest", func() {
 
 	var _ = Context("GetProposedTag", func() {
 		getProposedTag := func(latestTag, changelogDir, tag string) error {
--- a/changelogutils/changelog_test.go
+++ b/changelogutils/changelog_test.go
@@ -24,7 +24,7 @@ var _ = FDescribe("ChangelogTest", func() {
 		getProposedTag := func(latestTag, changelogDir, tag string) error {
 			fs := afero.NewOsFs()
 			actualTag, actualErr := changelogutils.GetProposedTag(fs, latestTag, changelogDir)
-			Expect(actualTag).To(BeEquivalentTo(tag), actualErr)
+			Expect(actualTag).To(BeEquivalentTo(tag), fmt.Sprintf("%v", actualErr))
 			return actualErr
 		}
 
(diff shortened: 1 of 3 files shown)


versionutils/version.go
```diff
@@ -125,17 +130,26 @@ func (v Version) IsGreaterThan(lesser Version) (bool, bool) {
...
 	}
 
-	if v.Label != lesser.Label {
+	if v.Label == lesser.Label {
+		if v.LabelVersion > lesser.LabelVersion {
+			return true, true
+		} else if v.LabelVersion < lesser.LabelVersion {
+			return false, true
+		}
 		return false, false
```
arianaw66, line 139: I think this case is actually the 2 versions are entirely identical? (Which used to be the end of the function.) So it would be false, true?

changelogutils/changelog_test.go
```diff
@@ -18,13 +18,13 @@ import (
 	"github.com/spf13/afero"
 )
 
-var _ = Describe("ChangelogTest", func() {
+var _ = FDescribe("ChangelogTest", func() {
```
arianaw66, line 21: Probably should add that "failOnFocus" flag or whatever it is.
```suggestion
var _ = Describe("ChangelogTest", func() {
```

versionutils/version.go
```diff
@@ -18,6 +18,11 @@ const (
 	SemverMinimumVersion  = "v0.0.1"
 )
 
+// SpeciallyOrderedPrefixes are prefixes that are ordered in a special way
+// This is exported so that consuming packages can specify their own special labels
+// These formats take precedence over alphanumeric ordering for labels
+var SpeciallyOrderedPrefixes = []string{"rc", "beta", "dev"}
```
arianaw66, line 24: I think this may be in reverse order? It's confusing to think about because "less" actually means lower priority, right? As a dev release will occur before a beta?
I believe we'd want: "-bogus0" < "-dev0" < "-beta0" < "-rc0".
So we need update the logic so that unrecognized prefixes are the least and then this list is reversed or considered as most to least.

arianaw66 (commented):
Thanks for taking a look at this!

versionutils/version.go
```diff
@@ -125,17 +130,26 @@ func (v Version) IsGreaterThan(lesser Version) (bool, bool) {
...
 		return false, false
 	}
 
-	if v.LabelVersion > lesser.LabelVersion {
-		return true, true
-	} else if v.LabelVersion < lesser.LabelVersion {
-		return false, true
+	// impose additional ordering based on our special labels
+	for _, label := range SpeciallyOrderedPrefixes {
+		if strings.HasPrefix(v.Label, label) {
```
arianaw66, line 144: Oh also I would probably lean toward exact match here. I think that would work and save us from complex regex code you mentioned in the PR description, unless I'm missing something.

versionutils/version.go
nfuden: This list as is will put priority on the first argument. Ie rc is latest

versionutils/version.go
nfuden: yeah! updated

nfuden:
probably should have been in the validator

versionutils/version.go
arianaw66: Added a unit test to confirm. Looks good.

versionutils/version.go
```diff
@@ -21,6 +21,7 @@ const (
 // SpeciallyOrderedPrefixes are prefixes that are ordered in a special way
 // This is exported so that consuming packages can specify their own special labels
 // These formats take precedence over alphanumeric ordering for labels
+// Labels take are ordered in increasing order of precedence
```
arianaw66, line 24: Isn't this... decreasing order of precedence? 
These take precedence over other labels, and "rc" takes precedence over "dev".

arianaw66 (commented):
Looks like this actually exists already: [IsGreaterThanWithLabelOrder](https://github.com/solo-io/go-utils/blob/v0.24.0/versionutils/version.go#L145)

PR #509, opened 2023-04-06
Remove actions/cache

Bslabe123:
See https://github.com/solo-io/gloo/pull/8062#issue-1656108774

### DIFF
--- a/.github/workflows/codeformat.yaml
+++ b/.github/workflows/codeformat.yaml
@@ -26,15 +26,15 @@ jobs:
 
           echo "After clearing disk space:"
           df -h
-      - name: Set up Go 1.18.1
-        uses: actions/setup-go@v4
-        with:
-          go-version: 1.18.1
-        id: go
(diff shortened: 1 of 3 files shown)


Bslabe123:
/kick https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=6aee4946-f8d8-4aa9-9acb-2532ba429a2b

.github/workflows/codeformat.yaml
```diff
@@ -27,20 +27,14 @@ jobs:
           echo "After clearing disk space:"
           df -h
       - name: Set up Go 1.18.1
-        uses: actions/setup-go@v3
+        uses: actions/setup-go@v4
         with:
           go-version: 1.18.1
```
jenshu, line 32: can you also switch both of these to use go-version-file?
you'll have to move the checkout before setup-go

sam-heilbron (approved):

Issue #511, opened 2023-05-02
Benchmark utils can report zero execution time


marcogschmidt:
The `TimeForFuncToComplete` function can fail [here](https://github.com/solo-io/go-utils/blob/248ace1ecbc843a4a99afe6e86274ed2f69891c5/testutils/benchmarking/benchmarking.go#L68) due to the measured runtime of the tested function being 0.

This is due to the fact that we are only taking `utime` into account, which is the time spent executing in user space. We should also account for time spent in kernel space, which could heavily impact the actual runtime of a function (e.g. due to I/O). See [this man page](https://man7.org/linux/man-pages/man2/getrusage.2.html) for more info about the `getrusage` syscall that we are relying on.

Occurrences of the problem:
- https://github.com/solo-io/ext-auth-service/actions/runs/4852953046/jobs/8650382165
- https://github.com/solo-io/ext-auth-service/actions/runs/4862562875/jobs/8669098496?pr=540
- https://github.com/solo-io/ext-auth-service/actions/runs/4862562875/jobs/8669246108?pr=540
- https://github.com/solo-io/ext-auth-service/actions/runs/4862562875/jobs/8669967003?pr=540

PR #512, opened 2023-05-02
New benchmark function

marcogschmidt:
## Summary
This PR adds the `Measure` function to the `benchmarking` test utility package. The new function measures the portion of time a given test function spent in user mode, in kernel mode, and in both of them combined. The other functions in the same package have been deprecated and will be removed when other repositories will stop depending on them.

## Motivation
The change was prompted by the fact that the current `TimeForFuncToComplete` function is in some instances reporting execution times of 0 nanoseconds, which causes failures in benchmark tests. I had encountered this problem [before](https://github.com/solo-io/ext-auth-service/pull/425#issuecomment-1276678054) and decided to spend some time to investigate it this time around as it seemed to be manifesting itself increasingly often during [my refactoring of CI in the ext-auth-service repository](https://github.com/solo-io/ext-auth-service/pull/540).

The issue is caused by the fact that we only measure time spent in user mode and that in some cases a function spends all of its time in kernel mode. An example can be found in [this test run](https://github.com/solo-io/ext-auth-service/actions/runs/4864169861/jobs/8672838054), where you can see that the `utime` (the first field in that poorly logged struct) is the same before and after the execution of the test function, while the `stime` (second field) goes from 0 to 15875µs: 

```
rusage1: {{0 149100} {0 0} 232368 0 0 0 78 0 0 0 0 0 0 0 15 61}
rusage2: {{0 149100} {0 15875} 232368 0 0 0 97 0 0 0 0 0 0 0 15 62}
```

I was not able to find out why this is happening in the time I gave myself to look into this. It would be interesting to understand this, but I don't think it is worth the effort at this moment.

## Solution
Regardless of the cause of the problem, I think we should take into account both the time we spend in user and kernel modes when benchmarking a function, so I added a new function that returns both of those pieces of information. This way the caller can decide which piece of information they want to use. I still kept a check on zero execution time, but it is run against `utime+stime` instead of just `utime`.

As a side note, the new function does not run assertions or write to standard output, I think that should be left to the caller as well.

BOT NOTES: 
resolves https://github.com/solo-io/go-utils/issues/511

### DIFF
--- a/.github/workflows/pull_request.yaml
+++ b/.github/workflows/pull_request.yaml
@@ -19,17 +19,17 @@ jobs:
     - uses: actions/checkout@v3
       with:
         fetch-depth: 0
-    - name: Setup GCloud Auth
-      uses: google-github-actions/auth@v1
+    - name: Set up Python
+      uses: actions/setup-python@v4
       with:
+        python-version: 3.8
+    - uses: google-github-actions/setup-gcloud@a48b55b3b0eeaf77b6e1384aab737fbefe2085ac
+      with:
+        version: '290.0.1'
         project_id: ${{ secrets.GCP_PROJECT_ID }}
-        credentials_json: ${{ secrets.GCP_SA_KEY }}
-        create_credentials_file: true
-        export_environment_variables: true
-    - name: Setup GCloud SDK
-      uses: google-github-actions/setup-gcloud@v1.1.0
-      with:
-        version: '>= 386.0.0'
+        service_account_key: ${{ secrets.GCP_SA_KEY }}
+        export_default_credentials: true
+      name: Gcloud Login
     - name: Install Trivy
       run: |
         wget https://github.com/aquasecurity/trivy/releases/download/v0.18.3/trivy_0.18.3_Linux-64bit.deb
--- a/testutils/benchmarking/benchmarking.go
+++ b/testutils/benchmarking/benchmarking.go
@@ -90,8 +90,11 @@ func TimeForFuncToComplete(f func()) float64 {
 	realRuntime := realDuration2 - realDuration1
 	fmt.Printf("realUserRuntime: %v\n", realRuntime)
 
-	fmt.Printf("utime: %f\n", userRuntime.Seconds())
-	Expect(userRuntime.Seconds()).Should(
+	fmt.Printf("utime:         %f\n", userRuntime.Seconds())
+	fmt.Printf("utime + stime: %f\n", realRuntime.Seconds())
+	Expect(realRuntime.Seconds()).Should(
 		BeNumerically(">", 0))
+
+	// TODO(marco): let's leave it as is so I don't need to adjust the thresholds for now
 	return userRuntime.Seconds()
 }
--- a/testutils/benchmarking/benchmarking.go
+++ b/testutils/benchmarking/benchmarking.go
@@ -87,7 +87,7 @@ type Result struct {
 }
 
 // Measure returns the time it took to execute the given function. It only compiles on Linux.
-// Most often you will want  to run the code you want to test in a loop to get more sizeable readings, e.g.:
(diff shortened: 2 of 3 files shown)


.github/workflows/pull_request.yaml
```diff
@@ -19,17 +19,17 @@ jobs:
...
-      uses: google-github-actions/setup-gcloud@v1.1.0
+        python-version: 3.8
+    - uses: google-github-actions/setup-gcloud@a48b55b3b0eeaf77b6e1384aab737fbefe2085ac
       with:
-        version: '>= 386.0.0'
+        version: '290.0.1'
+        project_id: ${{ secrets.GCP_PROJECT_ID }}
+        service_account_key: ${{ secrets.GCP_SA_KEY }}
+        export_default_credentials: true
+      name: Gcloud Login
```
marcogschmidt, line 32: This step was [failing](https://github.com/solo-io/go-utils/actions/runs/4866590886/jobs/8678320304) even though I have not changed anything. It seems to have worked on [the last PR against main](https://github.com/solo-io/go-utils/pull/509). I know this is not ideal as it pins us to an older version of the action, but my other attempts at fixing this failed and this works.

nfuden:
@sam-heilbron as you were looking into benchmarking. @ashishb-solo for visibility  as you had previously investigated the 0 count benchmark results

.github/workflows/pull_request.yaml
```diff
@@ -19,17 +19,17 @@ jobs:
...
-        credentials_json: ${{ secrets.GCP_SA_KEY }}
-        create_credentials_file: true
-        export_environment_variables: true
-    - name: Setup GCloud SDK
-      uses: google-github-actions/setup-gcloud@v1.1.0
+        python-version: 3.8
+    - uses: google-github-actions/setup-gcloud@a48b55b3b0eeaf77b6e1384aab737fbefe2085ac
       with:
-        version: '>= 386.0.0'
+        version: '290.0.1'
```
kdorosh, line 28: why use an older version?

kdorosh: ah i see https://github.com/solo-io/go-utils/pull/512/files#r1183133800

testutils/benchmarking/benchmarking.go
```diff
@@ -69,3 +75,73 @@ func TimeForFuncToComplete(f func()) float64 {
...
+		return Result{}, errors.New("total execution time was 0 ns")
+	}
+
+	return res, nil
+}
+
+func doMeasure(f func()) (before syscall.Rusage, after syscall.Rusage, err error) {
+
+	// Important: lock OS thread to ensure we are the only goroutine being benchmarked
+	runtime.LockOSThread()
```
ashishb-solo, line 128: does there need to be a corresponding unlock for this call?

.github/workflows/pull_request.yaml
marcogschmidt: See my [comment above](https://github.com/solo-io/go-utils/pull/512#discussion_r1183133800). The current action is failing unexpectedly with an error I have not seen before and which I could not solve quickly. I could not afford to spend more time to look into this, so I just used the same setup that we use in other repos currently and that I know works. Long term we should switch to using identity federation instead of relying on these API keys.

testutils/benchmarking/benchmarking.go
```diff
@@ -69,3 +75,73 @@ func TimeForFuncToComplete(f func()) float64 {
...
+
+	// Might just be paranoid, but disable garbage collection while benchmarking
+	prevGc := debug.SetGCPercent(-1)
+
+	defer func() {
+		debug.SetGCPercent(prevGc)
+		runtime.UnlockOSThread()
+	}()
```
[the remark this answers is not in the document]
kdorosh: nevermind, this is handled by your proposal, which is that `f()` itself has loops

kdorosh (approved):

ashishb-solo (approved):

PR #515, opened 2023-05-26
Helm indent finder

jackstine:
# Description

Includes `FindHelmChartWhiteSpaces()`, which takes a string, a yaml, and tries to parse it to find issues within the helm chart.  Primary this is trying to find issues with white spacing, where each line should be at most off by 0 or -2, given exceptions that work according to the Helm spec.  

The function will return a slice of windows (string slices) to represent the areas in the code that contain the issue that it found.

List of features
1. looks for spacing issues with the next line is off by more than 2
2. if next line is an array, it can be off by 4 or 2 spaces.
3. if the line has special breaks (`|-`, `>-`, etc.) identified in YAML, these are now ignored areas in the YAML.
4. if the YAML has empty lines, these are acceptable.
5. if the YAMl has spaces in an empty line, there is an option to control whether to see these or not. Use the `DetectWhiteSpacesInEmptyLines` option when parsing.

## Context

Users were having issues with formatting in the helm chart. Particularly when faced with this in the YAML. The 2nd line has 2 to many spaces, and so on.

```yaml
resources:
    name: gloo-role-knative
    labels:
        app: gloo
        gloo: rbac
```
## Use in Gloo

Link to [the Gloo PR](https://github.com/solo-io/gloo/pull/8302) on usage.


changelog/v0.24.3/helm-scanning.yaml
```diff
@@ -0,0 +1,6 @@
+changelog:
+- type: NON_USER_FACING
```
sam-heilbron, line 2: nit: Since this is an internal library, the "users" are more our devs teams. In that light, I almost view this as a new feature, one which didn't exist before and devlopers upgrading to this library version can use

sam-heilbron (commented):
I like this! And since it's entirely optional (ie not required or affeceting current code paths), Im going to rely on the output's from gloo/solo-projects to determine its accuracy.

If I'm a developer looking at this tool for the first time, it's unclear how I can use it. Can you add a readme which just outlines the basic usage, or perhaps highlight more clearly the funcitons that are intended to be called by other projects?

helm/helm_indent_finder.go
```diff
@@ -0,0 +1,198 @@
+package helm
```
sam-heilbron, line 1: most packages in this repo are named [package]utils. Can we update just to follow that paradigm?

helm/helm_indent_finder_test.go
```diff
@@ -0,0 +1,244 @@
...
+    - federatedratelimitconfigs
+    - federatedratelimitconfigs/status
+  verbs:
+    - '*'
+`
+			badWindows := helm.FindHelmChartWhiteSpaces(data, opts)
+			Expect(len(badWindows)).To(Equal(0))
+		})
+		It("should find that the array is off by one", func() {
+			// the line fed.ratelimit.solo.io is off by one space
```
sam-heilbron, line 155: Super helpful comment! I like it

helmutils/helm_indent_finder.go
```diff
@@ -0,0 +1,221 @@
+package helmutils
+
+import (
+	"strings"
+)
+
+/*
+	FindHelmChartWhiteSpaces is used to find unuasual spaces in helm charts, that could cause issues.
```
inFocus7, line 8:
```suggestion
	FindHelmChartWhiteSpaces is used to find unusual spaces in helm charts, that could cause issues.
```

helmutils/helm_indent_finder_test.go
```diff
@@ -0,0 +1,244 @@
...
+metadata:
+   labels:
+    app: gloo
+    gloo: rate-limit
+  name: rate-limit
+  namespace: default
+spec:
+`
+			badWindows := helmutils.FindHelmChartWhiteSpaces(data, opts)
+			Expect(len(badWindows)).To(Equal(2))
```
inFocus7, line 70: would it be worth to also do an Expect that the array of badWindows are the windows we expect?

helmutils/helm_indent_finder_test.go
```diff
@@ -0,0 +1,244 @@
...
+        runAsNonRoot: true
+        runAsUser: 10101
+`
+			badWindows := helmutils.FindHelmChartWhiteSpaces(data, opts)
+			Expect(len(badWindows)).To(Equal(0))
+		})
+	})
+
+	Describe("special breaks", func() {
+		DescribeTable("should accept special breaks", func(specialBreak string) {
```
inFocus7, line 216: What happens where there's special breaks _and_ more data afterward? For example, adding more fields/items after the `envoy.json` one

helmutils/helm_indent_finder_test.go
jackstine: don't think so.
Since the window size is relatively large compared to the input.
But I guess the statement should be, should we output the line that causes the issue, this can be up for debate.  We could return a structure that contains metadata about these windows.   

The reason I chose windows is that it was hard to identify the area in the code that was causing the problem. Especially if the code was re-used multiple times.

helmutils/helm_indent_finder_test.go
jackstine: not sure what you mean.

You have the following, in which the `data` and `more data ...` are just a part of the special break. Could you show me what you mean?
```yaml
something: |-
    data..
     more data...
 something1: info
 ```

File: helmutils/helm_indent_finder.go [hunk 7]
```diff
@@ -0,0 +1,221 @@
+package helmutils
+
+import (
+	"strings"
+)
+
+/*
```
sam-heilbron, line 7: niiiice!

helmutils/helm_indent_finder_test.go
inFocus7: I was thinking like
```
something: |-
  blah
     ijasd
  basu
someotherfield:
  - something
```

not sure if that's valid, but was wondering if we fully stopped checking the syntax after the special break, or if we continue checking different fields _outside_ that break. (if that makes sense)

jackstine: yeah this is covered.

`someotherfieldnotpartofabovebreak` will not cause an issue.
Updated the test, by adding this line.

inFocus7: Oh just for clarification before resolving, if that `someotherfieldblahblah` field had an error, it would be caught right?

jackstine: if `someotherfieldblahblah` was off it would not produce an error. This would be because it would considered the same block as the special break.  It would be read like this in YAML

jackstine: `someotherfieldblahblah` would have to start on the same indent or lower as the one that started the special brake.

sam-heilbron (approved):

PR #516, opened 2023-06-23
7299 logging to file

sheidkamp:
Adding the ability to log to a file instead of STDOUT by setting the LOG_TO_FILE_LOCATION environment variable. This is done to support AWS Fargate, which can not collect logs from STDOUT:

```
For pods running on Fargate, you need to use the sidecar pattern. You can run a Fluentd (or [Fluent Bit](http://fluentbit.io/)) sidecar container to capture logs produced by your applications. This option requires that the application writes logs to filesystem instead of stdout or stderr.
```
https://aws.amazon.com/blogs/containers/how-to-capture-application-logs-when-using-amazon-eks-on-aws-fargate/


contextutils/context_and_logging.go
```diff
@@ -42,6 +42,13 @@ const LogLevelEnvName = "LOG_LEVEL"
 func buildProductionLogger() (*zap.Logger, error) {
 	config := zap.NewProductionConfig()
 	config.EncoderConfig.EncodeTime = zapcore.ISO8601TimeEncoder
+
+	// For non-splt logging, allow the user/environment to specify a log file location.
+	// When we log to a file, we will not log it STDOUT.
+	if os.Getenv("LOG_TO_FILE_LOCATION") != "" {
```
elcasteel, line 48: nit: can we make `LOG_TO_FILE_LOCATION` a constant?

elcasteel:
Do we have plans to add this to documentation?

contextutils/context_and_logging.go
sheidkamp: good idea, done.

sheidkamp:
> Do we have plans to add this to documentation?

Great question @elcasteel - Looking a a similar piece of functionality (`SPLIT_LOG_OUTPUT`), the only documentation I can find is in the changelogs.

Happy to update appropriate documentation, if we have a spot.

elcasteel:
> > Do we have plans to add this to documentation?
> 
> Great question @elcasteel - Looking a a similar piece of functionality (`SPLIT_LOG_OUTPUT`), the only documentation I can find is in the changelogs.
> 
> Happy to update appropriate documentation, if we have a spot.

This is a bit of a gap for us right now. Perhaps we could add a note in a fargate related guide?

elcasteel (dismissed):

marcogschmidt (approved):

elcasteel (approved):

PR #523, opened 2023-11-23
chore: bump k8s to 1.28

davidjumani:
While this PR might not be necessary since there are no k8s dependencies, it would be nice to keep the dependencies around k8s v0.28 in sync with k8s-utils
This also modifies repo_client tests so that they do not hit the GH rate limit as seen in prior commit test failures


jenshu (commented):
can you update the PR description to explain why the ginkgo and other bumps are needed for k8s 1.28? (since the PR doesn't include a k8s bump)

changelog/v0.24.8/bump-k8s-128.yaml
```diff
@@ -0,0 +1,25 @@
...
+  dependencyRepo: ginkgo/v2
+  dependencyTag: v2.11.0
+- type: DEPENDENCY_BUMP
+  dependencyOwner: onsi
+  dependencyRepo: gomega
+  dependencyTag: v1.27.10
+- type: DEPENDENCY_BUMP
+  dependencyOwner: sirupsen
+  dependencyRepo: logrus
+  dependencyTag: v1.9.3
```
jenshu, line 18: looks like go.mod doesn't have this update

changelog/v0.24.8/bump-k8s-128.yaml
```diff
@@ -0,0 +1,25 @@
+changelog:
```
jenshu, line 1: can you also include the go 1.19 bump in changelog?

changelog/v0.24.8/bump-k8s-128.yaml
davidjumani: Fixed, Thanks!

jenshu (approved):

PR #524, opened 2023-12-08
[13501] Continue security scanning but highlight missing images

t-edris:
Previously, when an image for a specific version was unable to be found, the remaining images for that version would be skipped as a result. This behavior has been changed so a clear error message will be visible on the generated vulnerability report for that image and the remaining images will be scanned.

## Example usage of new flag

```
❯ cat developerDebugInstructions.md
## Debug Instructions

Use `make run-local-security-checks` to reproduce these results manually on your machine.

For quick iteration, choose an image to focus on (perhaps one with the majority of the CVEs) and run scans against it individually before ensuring the fixes have worked for all the images.

For example, use `make mgmt-server-image` to build the `mgmt-server` image and `./trivy image --severity HIGH,CRITICAL gcr.io/gloo-mesh/gloo-mesh-mgmt-server:${VERSION}` to scan it.

The UI image often has unique CVEs because it is built from separate Dockerfiles than the others. For this one, run `make ui-docker` and `./trivy image --severity HIGH,CRITICAL gcr.io/gloo-mesh/gloo-mesh-ui:${VERSION}`

```

```
❯ go run cli/main.go scan-repo -v -i ../../gloo-mesh-enterprise/docs/cmd/imageVersionConstraints.csv \
-c ">=v2.1.0-0" \
-r gcr.io/gloo-mesh \
-g gloo-mesh-enterprise \
-a none \
-d developerDebugInstructions.md
```

_Along with some temp code to output the final MD to a local file rather than a GH issue, just for this demo_

```go
func (g *GithubIssueWriter) CreateUpdateVulnerabilityIssue(. . .) error {
. . .
	f, err := os.Create("testGeneratedFile.md")
	if err != nil {
		return err
	}
	if _, err := f.WriteString(vulnerabilityMarkdown); err != nil {
		return err
	}
	if err := f.Sync(); err != nil {
		return err
	}
. . .
}
```

The generated file
Screenshot 2023-12-11 at 10 42 01 AM

also includes the new error
Screenshot 2023-12-11 at 10 42 20 AM

The expected usage of this would be in gloo-mesh-enterprise, similar to the usage of `docs/cmd/imageVersionConstraints.csv`.

BOT NOTES: 
resolves https://github.com/solo-io/gloo-mesh-enterprise/issues/13501


t-edris:
Unsure why changelog-bot expects v0.24.9. v0.24.8 is the latest release and there have been no commits to `main` since it has been released.

marcogschmidt:
> Unsure why changelog-bot expects v0.24.9. v0.24.8 is the latest release and there have been no commits to `main` since it has been released.
> 
> Maybe because the changelog doesn't have BREAKING_CHANGE?

I think that's it. The last release is [v0.24.8](https://github.com/solo-io/go-utils/tree/main/changelog/v0.24.8) and since you have marked this as FIX/NEW_FEATURE, it's expecting only a patch version bump

securityscanutils/securityscan.go
```diff
@@ -236,15 +241,21 @@ func (r *SecurityScanRepo) RunMarkdownScan(ctx context.Context, release *github.
...
 		}
 
 		if vulnFound {
 			trivyScanMd, err := ioutil.ReadFile(output)
 			if err != nil {
 				return eris.Wrapf(err, "error reading trivy markdown scan file %s to generate github issue", output)
 			}
-			vulnerabilityMd += fmt.Sprintf("# %s\n\n %s\n\n", imageWithRepo, trivyScanMd)
+			if r.Opts.DeveloperDebugInstructions != "" {
+				vulnerabilityMd += fmt.Sprintf("%s\n\n", r.Opts.DeveloperDebugInstructions)
```
SirNexus, line 256: Do you mind posting a screenshot of example usage here?

t-edris: Added to the PR description

changelog/v0.24.9/13501-report-missing-images-in-scan.yaml
```diff
@@ -0,0 +1,9 @@
+changelog:
+  - type: NEW_FEATURE
+    issueLink: https://github.com/solo-io/gloo-mesh-enterprise/issues/13501
```
arianaw66, line 3: I think this is actually from a different issue: https://github.com/solo-io/gloo-mesh-enterprise/issues/13499#issuecomment-1847675824

But does not resolve that issue, just the issue comment

securityscanutils/github_writer.go
```diff
@@ -83,6 +83,10 @@ func (g *GithubIssueWriter) CreateUpdateVulnerabilityIssue(ctx context.Context,
 		return nil
 	}
 
+	if developerDebugInstructions != "" {
+		vulnerabilityMarkdown = fmt.Sprintf("%s\n%s", developerDebugInstructions, vulnerabilityMarkdown)
```
arianaw66, line 87: Do we want "## Debug Instructions" to be added by go-utils rather than the calling repo? I could go either way.

arianaw66: But if the calling repo adds it, maybe the naming should be more generic for any additional context they want added to the GitHub issue. Actually, yeah, that's probably a better way to go in go-utils. Can we make the flag/var naming and description more general?

securityscanutils/trivy_scanner.go
```diff
@@ -16,6 +18,8 @@ import (
 // Status code returned by Trivy if a vulnerability is found
 const VulnerabilityFoundStatusCode = 52
 
+var ImageNotFoundError = errors.New("❗IMAGE MISSING UNEXPECTEDLY❗")
```
arianaw66, line 21: Will the emojis show up properly in GitHub? Do we need ":exclamation:" instead?

securityscanutils/trivy_scanner.go
```diff
@@ -85,7 +89,7 @@ func (t *TrivyScanner) executeScanWithRetries(ctx context.Context, scanArgs []st
 		}
 
 		// If there is no image, don't retry
-		if IsImageNotFoundErr(string(out)) {
+		if isImageNotFoundErr(string(out)) {
 			logger.Warnf("Trivy scan with args [%v] produced image not found error", scanArgs)
 
 			// swallow error if image is not found error, so that we can continue scanning releases
```
arianaw66, line 95: Probably needa update this comment

securityscanutils/trivy_scanner_test.go
```diff
@@ -56,7 +56,7 @@ var _ = Describe("Trivy Scanner", func() {
 		})
 		completed, vulnFound, err := t.ScanImage(context.TODO(), inputImage, inputMarkdownTemplateFile, outputFile)
 
-		Expect(err).NotTo(HaveOccurred())
+		Expect(err).To(MatchError(ImageNotFoundError))
```
arianaw66, line 59: Do we have any unit tests that check more than 1 image at a time to ensure it's continuing after the error?

arianaw66 (commented):
A few tweaks, but looks good overall!

changelog/v0.24.9/13501-report-missing-images-in-scan.yaml
t-edris: Good catch, thank you!

securityscanutils/github_writer.go
t-edris: Yeah good suggestion, I'll update

securityscanutils/trivy_scanner.go
t-edris: https://github.com/t-edris/testing-things

It looks like the actual emoji's do show up and `:exclamation:` does in some cases. I don't understand the rules for that quite yet as evidenced by the commit history there.

It seems most consistent to keep this as-is, since that seems to always render.

securityscanutils/trivy_scanner_test.go
t-edris: I added a missing image name to an existing test suite that asserts the generated files for all images exist after the scan completes.

securityscanutils/securityscan.go
```diff
@@ -233,10 +248,13 @@ func (r *SecurityScanRepo) RunMarkdownScan(ctx context.Context, release *github.
 			imageWithRepo = fmt.Sprintf("%s/%s:%s", r.Opts.ImageRepo, image, version)
 		}
 		fileName := fmt.Sprintf("%s_cve_report.docgen", image)
-		output := path.Join(outputDir, fileName)
+		output := path.Join(trivyScanOutputDir, fileName)
 		_, vulnFound, err := r.trivyScanner.ScanImage(ctx, imageWithRepo, markdownTplFile, output)
 		if err != nil {
-			return eris.Wrapf(err, "error running image scan on image %s", imageWithRepo)
+			if !errors.Is(err, ImageNotFoundError) {
```
sam-heilbron, line 254: Will this ever be false? https://github.com/solo-io/go-utils/blob/main/securityscanutils/trivy_scanner.go#L91 we have some logic to handle missing errors, and we "swallow" the error.

securityscanutils/securityscan.go
```diff
@@ -248,8 +266,14 @@ func (r *SecurityScanRepo) RunMarkdownScan(ctx context.Context, release *github.
...
 	}
+	localOutputFilename := ""
+	if r.Opts.OutputResultLocally {
+		githubIssueOutputDir := path.Join(r.Opts.OutputDir, r.Repo, "github_issue_results")
+		err = os.MkdirAll(githubIssueOutputDir, os.ModePerm)
+		localOutputFilename = path.Join(githubIssueOutputDir, version+".md")
+	}
 	// Create / Update Github issue for the repo if a vulnerability is found
-	return r.githubIssueWriter.CreateUpdateVulnerabilityIssue(ctx, release, vulnerabilityMd)
+	return r.githubIssueWriter.CreateUpdateVulnerabilityIssue(ctx, release, vulnerabilityMd, r.Opts.AdditionalContext, localOutputFilename)
```
sam-heilbron, line 276: Should this always be run? Now that the "vulnerablityAction" has 3 options:  "output-locally", "github-issue-all", "github-issue-latest", shouldn't we _only_ use the Github writer if `github-issue-all` or `github-issue-latest` is set?

securityscanutils/trivy_scanner.go
```diff
@@ -104,6 +105,6 @@ func (t *TrivyScanner) executeScanWithRetries(ctx context.Context, scanArgs []st
 	return false, false, eris.Errorf("Trivy scan with args [%v] did not complete after %d attempts", scanArgs, t.scanMaxRetries)
 }
 
-func IsImageNotFoundErr(logs string) bool {
```
sam-heilbron, line 107: Technically, since we are changing the signature of an exported method, this would make these changes breaking. It could be that this method was never consumed outside of this package, so it might not be an issue, I just wanted to call it out as something to be on the lookout moving forward

securityscanutils/commands/utils.go
```diff
@@ -45,3 +45,11 @@ func GetImagesPerVersionFromFile(constraintsFile string) (map[string][]string, e
 
 	return imagesPerVersion, nil
 }
+
+func GetAdditionalContextFromFile(filename string) (string, error) {
```
sam-heilbron, line 49: Given that this function is generic and could be re-used, do you think it would be valuable to define this in `fileutils`, something like:
```
func ReadFileString(filename string)(string, error)
```

securityscanutils/securityscan.go
sam-heilbron: Ahh I see now when I looked lower at the code. I like the change? 

One thought - it seems we have two types of errors now: unrecoverableError (one that should cause us to short circuit), recoverableErr (one that we should continue the code execution). This would allow us to not require consumers of the `ScanImage` API, to be aware of the underlying reasons (that the image wasn't found), just whether or not the err should be recoverable or not.

not a requirement to implement, just figured I'd share the thought

securityscanutils/securityscan.go
```diff
@@ -248,8 +266,14 @@ func (r *SecurityScanRepo) RunMarkdownScan(ctx context.Context, release *github.
 		}
 
 	}
+	localOutputFilename := ""
+	if r.Opts.OutputResultLocally {
+		githubIssueOutputDir := path.Join(r.Opts.OutputDir, r.Repo, "github_issue_results")
+		err = os.MkdirAll(githubIssueOutputDir, os.ModePerm)
```
sam-heilbron, line 272: These two lines seem to be responsible for setting up the output directory, and I imagine they only need to be executed once. At the moment, they are executed for each version that is scanned, which means that if we scan 100 images (often more), we will execute this 100 times instead of just once.

I think we can move this setup logic high in the function to just run once, what do you think?

sam-heilbron (commented):
I like the changes, and really appreciate the thorough PR body that outline the change set. I had a few nits (can be ignored) and a few questions. If it's easier to go over the questions in person, feel free to toss time on my calendar or we can join a huddle

securityscanutils/trivy_scanner.go
t-edris: I think that's a good callout. My assumption here was that only the `securityscanutils` package was using this function, but I technically can't guarantee that nothing outside `go-utils` is using this, so I'll revert this change.

securityscanutils/predicate.go
```diff
@@ -11,7 +11,7 @@ import (
 // The securityScanRepositoryReleasePredicate is responsible for defining which
 // github.RepositoryRelease artifacts should be included in the bulk security scan
 // At the moment, the two requirements are that:
-// 1. The release is not a pre-release or draft
```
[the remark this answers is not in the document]
t-edris: The rationale here is to attempt to catch vulnerabilities before we GA a release version. Looking at the Changelog I added, I do see that this is not specified, so I'll add that as `NEW_FEATURE` (or would it be `BREAKING_CHANGE` in your opinion?)

From https://github.com/solo-io/gloo-mesh-enterprise/issues/13499

>Currently, we only generate security alert issues like https://github.com/solo-io/gloo-mesh-enterprise/issues/13151 for stable branches, as you can see in the [logs](https://github.com/solo-io/gloo-mesh-enterprise/actions/runs/7087689707/job/19288523223) of the security-scan-scheduled job.

> This means we won't catch vulnerabilities on main until after GA release. We should be catching these as soon as possible, so for betas and rcs as well.

securityscanutils/github_writer.go
```diff
@@ -74,7 +75,13 @@ func (g *GithubIssueWriter) getAllGithubIssues(ctx context.Context) ([]*github.I
 // Creates/Updates a Github Issue per image
 // The github issue will have the markdown table report of the image's vulnerabilities
 // example: https://github.com/solo-io/solo-projects/issues/2458
-func (g *GithubIssueWriter) CreateUpdateVulnerabilityIssue(ctx context.Context, release *github.RepositoryRelease, vulnerabilityMarkdown string) error {
+func (g *GithubIssueWriter) CreateUpdateVulnerabilityIssue(
+	ctx context.Context,
+	release *github.RepositoryRelease,
+	vulnerabilityMarkdown string,
+	developerDebugInstructions string,
+	outputLocalFilename string,
```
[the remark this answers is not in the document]
t-edris: I thought this was a good suggestion, so I refactored with a new `IssueWriter` interface with a few implementations. Since I created an `issuewriter` package and moved the `github_writer` implementation into it, I updated the changelog to specify `BREAKING_CHANGE`.

securityscanutils/commands/scan_repo.go
```diff
@@ -70,6 +75,10 @@ func doScanRepo(ctx context.Context, opts *scanRepoOptions) error {
 	if err != nil {
 		return err
 	}
+	additionalContext, err := fileutils.ReadFileString(opts.additionalContextFile)
```
sam-heilbron, line 78: How does this behave when the additionalContextFile is empty, which is the default behavior? I think it might be helpful to have a simple unit test on this utility which shows the behaviors.

securityscanutils/issuewriter/issue_writer.go
```diff
@@ -0,0 +1,11 @@
+package issuewriter
+
+import (
+	"context"
+
+	"github.com/google/go-github/v32/github"
+)
+
+type IssueWriter interface {
```
sam-heilbron, line 9: I like this interface, and think that it makes a lot of sense in the context of this PR. Since the idea is fresh, I know how this operates, but I'm sure in 1 month, or 6 months, I'll forget the contract of this Interface. Could you add a comment that clarifies how this is used?

I think also clarifying what each parameter of the Write function is, would be useful.

securityscanutils/issuewriter/local_writer.go
```diff
@@ -0,0 +1,47 @@
...
+	"context"
+	"fmt"
+	"os"
+	"path"
+
+	"github.com/Masterminds/semver/v3"
+	"github.com/google/go-github/v32/github"
+)
+
+type LocalIssueWriter struct {
```
sam-heilbron, line 13: nit: Since this is exported, I think a comment clarifying what a local_writer is, and how it is intended to be used, would be valuable

securityscanutils/issuewriter/local_writer.go
```diff
@@ -0,0 +1,47 @@
...
+	filename := path.Join(l.outputDir, version.String()+".md")
+	f, err := os.OpenFile(filename, os.O_CREATE|os.O_APPEND|os.O_WRONLY, 0644)
+	if err != nil {
+		return err
+	}
+	_, err = fmt.Fprintf(f, contents)
+	if err != nil {
+		return err
+	}
+	return nil
```
sam-heilbron, line 46: What if we just conslidate this to? (happy to leave as is, also):
```suggestion
	return err
```

securityscanutils/issuewriter/local_writer.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/Masterminds/semver/v3"
+	"github.com/google/go-github/v32/github"
+)
+
+type LocalIssueWriter struct {
+	repo      GithubRepo
+	outputDir string
+}
+
+var liw IssueWriter = &LocalIssueWriter{}
```
sam-heilbron, line 18: Given that this is a new implementation of an interface, I think it would be helpful to have a simple suite of unit tests that define the contract of this struct. Tests like:
- Can write to file for validate release
- Writes to separate files for subsequent releases
- Appends contents to file, if one already exists (is this the desired behavior)?

securityscanutils/issuewriter/noop_writer.go
```diff
@@ -0,0 +1,19 @@
+package issuewriter
+
+import (
+	"context"
+
+	"github.com/google/go-github/v32/github"
+)
+
+type NoopWriter struct{}
```
sam-heilbron, line 9: I really like the null object pattern!

securityscanutils/predicate.go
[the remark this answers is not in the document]
sam-heilbron: I could be misunderstanding something here, so let me clarify my understand. A Github release has 3 properties that are relevant to use:
- [draft (bool)](https://github.com/google/go-github/blob/fb8a83de3e7046c49ab76e09a4e9b5857f8f3997/github/repos_releases.go#L26)
- [pre-release (bool)](https://github.com/google/go-github/blob/fb8a83de3e7046c49ab76e09a4e9b5857f8f3997/github/repos_releases.go#L27)
- [tag (string)](https://github.com/google/go-github/blob/fb8a83de3e7046c49ab76e09a4e9b5857f8f3997/github/repos_releases.go#L22)

I have always interpreted releases that are "pre-release" to be a flag that is useful for "deactivating" a release. That is, the pre-release state, is independent of the tag.

What you're getting at is that you want tags that are beta releases to be supported. My understanding was that your request is unreleated to pre-release, though I could be entirely wrong.

I would have expected your solution to define a semver.Constraint that supports beta releases. Did that alone not work?

securityscanutils/securityscan.go
```diff
@@ -233,23 +259,29 @@ func (r *SecurityScanRepo) RunMarkdownScan(ctx context.Context, release *github.
...
 			}
 			vulnerabilityMd += fmt.Sprintf("# %s\n\n %s\n\n", imageWithRepo, trivyScanMd)
 		}
 
 	}
-	// Create / Update Github issue for the repo if a vulnerability is found
-	return r.githubIssueWriter.CreateUpdateVulnerabilityIssue(ctx, release, vulnerabilityMd)
+	if vulnerabilityMd != "" && r.Opts.AdditionalContext != "" {
+		vulnerabilityMd = fmt.Sprintf("%s\n%s", r.Opts.AdditionalContext, vulnerabilityMd)
+	}
```
sam-heilbron, line 282: I like this! Perhaps just a clarifying question: is the additional context intended to just be used for the localWriter, or should it apply to all implementations?

securityscanutils/securityscan.go
```diff
@@ -233,23 +259,29 @@ func (r *SecurityScanRepo) RunMarkdownScan(ctx context.Context, release *github.
...
 		fileName := fmt.Sprintf("%s_cve_report.docgen", image)
-		output := path.Join(outputDir, fileName)
+		output := path.Join(trivyScanOutputDir, fileName)
 		_, vulnFound, err := r.trivyScanner.ScanImage(ctx, imageWithRepo, markdownTplFile, output)
 		if err != nil {
-			return eris.Wrapf(err, "error running image scan on image %s", imageWithRepo)
+			if errors.Is(err, UnrecoverableErr) {
+				return eris.Wrapf(err, "error running image scan on image %s", imageWithRepo)
+			}
+			vulnerabilityMd += fmt.Sprintf("# %s\n\n %s", imageWithRepo, err)
```
sam-heilbron, line 268: Could you point me to the test that shows this working in both the positive (unrecoverable error is short-circuited) and negative (recoverable error continues)

sam-heilbron (commented):
This is great! One part of this that I think is so powerful, is that we can now write a full suite of e2e tests that persist scan outputs into local files, whereas previously we were tightly coupled to GitHub. Thanks for removing that coupling.

The main open question I have is around my own undersatnding of a release being considered "pre-release".

Apart from that, just some small requests for enhanced tests. Since this is used so widely within our codebase, and is the source of truth for identifying vulnerabilities that impact our customers in production, I think having a lot of safe guards for this code will save us headaches in the future.

securityscanutils/trivy_scanner.go
arianaw66: Nice exploration, thanks for checking it out! Sounds good.

changelog/v0.25.0/13501-report-missing-images-in-scan.yaml
```diff
@@ -0,0 +1,17 @@
...
+    description: >
+      Scanner includes an optional Developer Debug Instructions section on reported vulnerabilities.
+  - type: NEW_FEATURE
+    issueLink: https://github.com/solo-io/gloo-mesh-enterprise/issues/13499
+    description: >
+      Scanner evaluates pre-release images, such as `*-beta1` and `*-rc2`.
+  - type: BREAKING_CHANGE
+    issueLink: https://github.com/solo-io/gloo-mesh-enterprise/issues/13499
+    description: >
+      Create `IssueWriter` interface with implementations as `NoopWriter`, `GithubIssueWriter`, and `LocalIssueWriter`.
```
arianaw66, line 13:
```suggestion
    description: >
      Scanner includes an optional Additional Context section on reported vulnerabilities.
    resolvesIssue: false
  - type: NEW_FEATURE
    issueLink: https://github.com/solo-io/gloo-mesh-enterprise/issues/13499
    description: >
      Scanner evaluates pre-release images, such as `*-beta1` and `*-rc2`.
    resolvesIssue: false
  - type: BREAKING_CHANGE
    issueLink: https://github.com/solo-io/gloo-mesh-enterprise/issues/13499
    description: >
      Create `IssueWriter` interface with implementations as `NoopWriter`, `GithubIssueWriter`, and `LocalIssueWriter`.
    resolvesIssue: false
```

securityscanutils/trivy_scanner_test.go
arianaw66: Nice. If we can confirm the generated files contain the later images "gloo" and "discovery", that would be nice, but I'm not familiar enough with the test suite to know how much of a pain that would be.

securityscanutils/predicate.go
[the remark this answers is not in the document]
arianaw66: Ah okay, I think I see what's happening here. The term pre-release is overloaded. Semver considers rc and beta tags as pre-releases. But @t-edris, you saw this go-utils library run for beta releases without making the code change on line 27 that removes `if release.GetPrerelease()`, right?
So we probably do not need any functional change here and rather should clarify with comments that `release.GetPrerelease()` is referring to a pre-release in the GitHub sense as Sam supposed and not in the Semver sense as I did.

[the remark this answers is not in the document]
arianaw66: Ah okay, I think I see what's happening here. The term pre-release is overloaded. Semver considers rc and beta tags as pre-releases. But @t-edris, you saw this go-utils library run for beta releases without making the code change on line 27 that removes `if release.GetPrerelease()`, right?
So we probably do not need any functional change here and rather should clarify with comments that `release.GetPrerelease()` is referring to a pre-release in the GitHub sense as Sam supposed and not in the Semver sense as I did.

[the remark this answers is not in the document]
arianaw66: And accordingly, we can leave the prerelease test as it was and add the new tests for rc and betas but without the `Prerelease: github.Bool(true),` lines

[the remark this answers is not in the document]
t-edris: I'll add a clarifying comment, thanks for the help with understanding this!

securityscanutils/trivy_scanner_test.go
t-edris: For reference, this is covered by this assertion in `securityscan_test.go`

```go
ExpectDirToHaveFiles(path.Join(markdownDir, "1.7.0"), "discovery_cve_report.docgen", "gloo_cve_report.docgen")
```

The `.docgen` files will only exist if they were scanned.

securityscanutils/issuewriter/local_writer.go
t-edris: I typically use this pattern for consistency and to be fully explicit and avoid mixing return patterns. In writing this PR, I almost introduced an issue (luckily caught by a unit test) where I mistakenly wrapped a `return err` with `return eris.Wrapf(MyError, err.Error())` because I saw a `return err`, but that `err` value was actually `nil`.

securityscanutils/securityscan.go
t-edris: The intention would be to apply to all implementations

securityscanutils/securityscan.go
t-edris: Adding additional tests to `securityscan_test.go` in an upcoming commit for this

securityscanutils/issuewriter/local_writer.go
```diff
@@ -0,0 +1,47 @@
...
+	"fmt"
+	"os"
+	"path"
+
+	"github.com/Masterminds/semver/v3"
+	"github.com/google/go-github/v32/github"
+)
+
+type LocalIssueWriter struct {
+	repo      GithubRepo
```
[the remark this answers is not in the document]
t-edris: Copy-paste remnants, we don't need it 👍

securityscanutils/predicate_test.go
```diff
@@ -34,9 +34,6 @@ var _ = Describe("Predicate", func() {
 			Entry("release is draft", &github.RepositoryRelease{
 				Draft: github.Bool(true),
 			}, false),
-			Entry("release is pre-release", &github.RepositoryRelease{
```
sam-heilbron, line 37: I think we still need this test, right? This test is used to demonstrate that github pre-releases are included

securityscanutils/securityscan.go
```diff
@@ -200,7 +217,16 @@ func (s *SecurityScanner) initializeRepoConfiguration(ctx context.Context, repo
...
+	if useGithubWriter {
+		repo.issueWriter = issuewriter.NewGithubIssueWriter(githubRepo, s.githubClient, issuePredicate)
+	} else if repo.Opts.OutputResultLocally {
+		repo.issueWriter, err = issuewriter.NewLocalIssueWriter(path.Join(repo.Opts.OutputDir, githubRepo.RepoName, "issue_results"))
+		if err != nil {
+			return err
+		}
+	} else {
+		repo.issueWriter = issuewriter.NewNoopWriter()
+	}
```
sam-heilbron, line 229: The log line below this file is no longer accurate. Now that we don't always generate GithubIssueWriter this will need to be adjusted

sam-heilbron (commented):
LGTM! One small comment about a test case that I think needs to be re-added

sam-heilbron (commented):
Really thanks for this thorough work! Since this repo only requires 1 approval, I am not going to approve the PR since there are open threads/discussions with other Engineers. However leaving this comment to signify that this LGTM

securityscanutils/trivy_scanner_test.go
arianaw66: The `.docgen` files will only exist is _something_ was scanned, right? But having "gloo" and "discovery" specifically confirms that it continues after "thisimagecannotbefound", unless I misunderstood.

Though seeing the recent update to the v2.4.6 Security Alert issue, I'm wondering if order isn't maintained anyway.

arianaw66 (approved):
I haven't done a thorough review of the more recent updates, but my comments have been sufficiently resolved and it looks good from a quick-pass; I'm approving per the advice of counsel (i.e. https://github.com/solo-io/go-utils/pull/524#pullrequestreview-1787352991 - I think I was the only one with outstanding comments)

PR #526, opened 2024-05-23
BuildLoggerFromConfig

bewebi:
Add `BuildLoggerFromConfig()` function that allows a caller to create a logger using a `zap.Config`

The key functionality this adds a way to create a new logger that respects the `LOG_TO_FILE_LOCATION` env var without the caller having to have knowledge of that env var

This is used in solo-io/dev-portal#2893 [here](https://github.com/solo-io/dev-portal/pull/2893/files#diff-f36fcc143c1388c7c89b28b165cfb74d7f4c00fd6e58667a5594437f13180436R83) to replace an instance where a new logger was previously created directly using the `zaputil.NewRaw()` function

An alternative would be to have consumers leverage the `LogToFileLocation` const to determine the output location, create a `zap.Config` and call `config.Build()` directly
Since this also requires callers to depend on `contextutils`, we may as well have the util library create the logger, handling logic for file location and default timestamp format

### DIFF
--- a/contextutils/context_and_logging.go
+++ b/contextutils/context_and_logging.go
@@ -109,16 +109,19 @@ func init() {
 		buildLogger = buildSplitOutputProductionLogger
 	}
 	if logger, err := buildLogger(); err != nil {
+		fmt.Printf("ERROR BUILDING LOGGER: %v\n", err)
 
 		// We failed to create a fallback logger. Our fallback
 		// unfortunately falls back to noop.
 		fallbackLogger = zap.NewNop().Sugar()
 	} else {
+		fmt.Printf("SETTING FALLBACKLOGGER TO LOGGER FROM BUILDLOGGER()\n")
 		fallbackLogger = logger.Sugar()
 	}
 }
 
 func SetFallbackLogger(logger *zap.SugaredLogger) {
+	fmt.Printf("IN SetFallbackLogger()\n")
 	fallbackLogger = logger
 }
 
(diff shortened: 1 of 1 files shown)


contextutils/context_and_logging.go
```diff
@@ -114,6 +114,22 @@ func init() {
 	}
 }
 
+func BuildLoggerFromConfig(config zap.Config) (*zap.Logger, error) {
```
nfuden, line 117: Why are we only respecting certain configuration items? Shouldnt we just be adding to the existing background logger to make sure that things respect previously set settings?

This feels like a bandaid to a larger problem instead of addressing how we pull from background in a non-context dependant way

bewebi: The whole config is respected, as we return the output from `config.Build()` 

This covers a case [here](https://github.com/solo-io/dev-portal/pull/2893/files#diff-f36fcc143c1388c7c89b28b165cfb74d7f4c00fd6e58667a5594437f13180436L85) where we were previously creating a brand new logger directly from a `zap.Config`, but not incorporating the `LOG_TO_FILE_LOCATION` env

If we don't want to directly support creating a logger from config as part of this util we can find another path forward
There isn't a great reason not to other than that having the option may be confusing since in most cases it's preferable to use the logger from context

nfuden: Why arent we pulling from our fallback logger?
This seems like its still a mismatch in where one can configure logging and how it may or may not be respected

bewebi:
https://github.com/solo-io/dev-portal/pull/2893 merged without need for this change

PR #531, opened 2024-08-21
Update dependencies where possible and move to go1.23

tjons:
This PR upgrades the language to go1.23 and upgrades dependencies to the latest versions used by upstream k8s, where possible. Our long list of dependencies seemed like it hadn't been checked in a while so I brought our versions to parity.

As we walk through the broader dependency upgrade project, we can catch issues earlier if we start introducing new dependency versions as soon as possible, instead of upgrading them in a dependant repository.


sam-heilbron (commented):
Should we make this a minor version bump to denote to consumers of the library that there are potentially large changes, with changing functionality? (Side effect too, is that if we need to backport work to LTS branches using earlier versions of Go,k8s it will be easier)

tjons:
@sam-heilbron I thought of that and did a little digging. I didn't see that done in the past in this repo for dependency upgrades without interface changes, but I think it makes a ton of sense. Let's make this `v0.26.0`!

sam-heilbron (approved):

jenshu (approved):

PR #563, opened 2026-04-28
Drop unneeded patch version in go mod

howardjohn:
https://blog.howardjohn.info/posts/go-mod-version/

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/solo-io/go-utils
 
-go 1.26.1
+go 1.26.0
 
 require (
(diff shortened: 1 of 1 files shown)


nfuden (approved):

nfuden:
If we want to release with just this then we need a changelog I think?

Also it might be nice to have something that just checks the go mod and enforces this so we dont have a push pull

howardjohn:
idk how to mark this as "no changelog needed"

nfuden:
Downside is it may break release depending on how it works in this repo now XD

